PR #32, opened 2022-03-03
feat: generalize simulation to all wallets

bhgomes:
Generalizes simulation for all wallets of the form `Wallet<C, L, S>` for where `C` is the transfer configuration, `L` is the ledger connection, and `S` is the signer connection. This allows for testing with more interaction complexity like substrate in-memory ledger, full node ledger, full signer, etc.

To build your own simulation see how the `manta-pay/src/simulation/mod.rs` file runs the new generalized simulation and copy the overall pattern.


bhgomes:
Closed in favor of #33.

Issue #24, opened 2022-02-27
Use Pre-computed Powers in Scalar Multiplication


bhgomes:
Right now we use a slow kind of scalar multiplication in-circuit where we compute powers of the base while multiplying. We can pre-compute these powers before entering the circuit and save on a lot of constraints. We can leverage the `PrecomputedScalarMul` trait and the other traits in `manta_crypto::ecc` to do this.

Since this is a parameter change, we need to make sure to upgrade the SDK whenever we add this in.

tsunrise:
What's the purpopse of `PreprocessedScalarMul` in current codebase?

SupremoUGH:
https://github.com/Manta-Network/manta-rs/pull/196 closes this. The corresponding trait is `FixedBaseScalarMul`.

PR #34, opened 2022-03-08
Support scalar mul of arkworks curve using precomputed table

tsunrise:
Adds support for scalar multiplication using precomputed bases. 

Closes: #24

### DIFF
--- a/manta-crypto/src/ecc.rs
+++ b/manta-crypto/src/ecc.rs
@@ -165,7 +165,9 @@ where
 /// Elliptic Curve Group
 pub trait Group<COM = ()>: PointAdd<COM> + PointDouble<COM> + ScalarMul<COM> {}
 
-/// Pre-processed Scalar Multiplication Table, represented as powers of two of `scalar`.
+/// Pre-processed Scalar Multiplication Table.
+///
+/// This table contains power-of-two multiples of a fixed base group element.
 #[cfg_attr(
     feature = "serde",
     derive(Deserialize, Serialize),
--- a/manta-crypto/src/ecc.rs
+++ b/manta-crypto/src/ecc.rs
@@ -138,8 +138,8 @@ pub trait ScalarMul<COM = ()> {
 pub trait PreprocessedScalarMul<COM, const N: usize>: ScalarMul<COM> + Sized {
     /// Performs the scalar multiplication against a pre-computed table.
     ///
-    /// The pre-computed table is a list of power-of-two multiples of `scalar`, such that
-    /// `table[i] = scalar * 2^i`.
+    /// The pre-computed table is powers of two of `scalar`, such that
+    /// `table[i] = 2^i * G`.
     #[must_use]
     fn preprocessed_scalar_mul(
         table: &[Self; N],
@@ -165,7 +165,7 @@ where
 /// Elliptic Curve Group
 pub trait Group<COM = ()>: PointAdd<COM> + PointDouble<COM> + ScalarMul<COM> {}
 
-/// Pre-processed Scalar Multiplication Table
+/// Pre-processed Scalar Multiplication Table, represented as powers of two of `scalar`.
 #[cfg_attr(
     feature = "serde",
     derive(Deserialize, Serialize),
@@ -183,17 +183,16 @@ pub struct PreprocessedScalarMulTable<G, const N: usize> {
 }
 
 impl<G, const N: usize> PreprocessedScalarMulTable<G, N> {
-    /// Builds a new [`PreprocessedScalarMulTable`] collection from `base`.
+    /// Builds a new [`PreprocessedScalarMulTable`] collection from `base`, such that `table[i] = 2^i * base`.
     #[inline]
     pub fn from_base<COM>(mut base: G, compiler: &mut COM) -> Self
     where
         G: Clone + PointAdd<COM, Output = G> + PointDouble<COM, Output = G>,
     {
         let mut powers = Vec::with_capacity(N);
-        let double = base.double(compiler);
         for _ in 0..N {
             powers.push(base.clone());
-            base.add_assign(&double, compiler);
+            base.double_assign(compiler);
         }
         Self::from_powers_unchecked(
             powers
--- a/manta-crypto/src/ecc.rs
+++ b/manta-crypto/src/ecc.rs
@@ -145,20 +145,6 @@ pub trait PreprocessedScalarMul<COM, const N: usize>: ScalarMul<COM> + Sized {
     ) -> Self::Output;
 }
 
-impl<G, COM> PreprocessedScalarMul<COM, 1> for G
-where
-    G: ScalarMul<COM>,
-{
-    #[inline]
-    fn preprocessed_scalar_mul(
-        table: &[Self; 1],
-        scalar: &Self::Scalar,
-        compiler: &mut COM,
-    ) -> Self::Output {
-        table[0].scalar_mul(scalar, compiler)
-    }
-}
-
 /// Elliptic Curve Group
 pub trait Group<COM = ()>: PointAdd<COM> + PointDouble<COM> + ScalarMul<COM> {}
 
--- a/manta-crypto/src/ecc.rs
+++ b/manta-crypto/src/ecc.rs
@@ -137,12 +137,29 @@ pub trait ScalarMul<COM = ()> {
 /// Elliptic Curve Pre-processed Scalar Multiplication Operation
 pub trait PreprocessedScalarMul<COM, const N: usize>: ScalarMul<COM> + Sized {
     /// Performs the scalar multiplication against a pre-computed table.
+    ///
+    /// The pre-computed table is a list of power-of-two multiples of `scalar`, such that
+    /// `table[i] = scalar * 2^i`.
     #[must_use]
     fn preprocessed_scalar_mul(
         table: &[Self; N],
         scalar: &Self::Scalar,
         compiler: &mut COM,
     ) -> Self::Output;
+}
+
+impl<G, COM> PreprocessedScalarMul<COM, 1> for G
+where
+    G: ScalarMul<COM>,
+{
+    #[inline]
+    fn preprocessed_scalar_mul(
+        table: &[Self; 1],
+        scalar: &Self::Scalar,
+        compiler: &mut COM,
+    ) -> Self::Output {
+        table[0].scalar_mul(scalar, compiler)
+    }
 }
 
 /// Elliptic Curve Group
--- a/manta-pay/Cargo.toml
+++ b/manta-pay/Cargo.toml
@@ -78,7 +78,7 @@ simulation = [
 std = ["manta-accounting/std", "manta-util/std"]
 
 # Testing Frameworks
-test = ["manta-accounting/test", "manta-crypto/test"]
+test = ["manta-accounting/test", "manta-crypto/test", "manta-crypto/getrandom"]
 
 # Wallet
 wallet = ["bip32", "manta-crypto/getrandom", "std"]
--- a/manta-pay/Cargo.toml
+++ b/manta-pay/Cargo.toml
@@ -78,7 +78,7 @@ simulation = [
 std = ["manta-accounting/std", "manta-util/std"]
 
 # Testing Frameworks
-test = ["manta-accounting/test", "manta-crypto/test", "manta-crypto/getrandom"]
+test = ["manta-accounting/test", "manta-crypto/test"]
 
 # Wallet
 wallet = ["bip32", "manta-crypto/getrandom", "std"]
@@ -120,3 +120,4 @@ tungstenite = { version = "0.17.2", optional = true, default-features = false, f
 
 [dev-dependencies]
 manta-pay = { path = ".", features = ["groth16", "test"] }
+manta-crypto = { path = "../manta-crypto", default-features = false, features = ["getrandom"] }
--- a/manta-pay/src/crypto/ecc/arkworks.rs
+++ b/manta-pay/src/crypto/ecc/arkworks.rs
@@ -23,11 +23,9 @@ use ark_r1cs_std::ToBitsGadget;
 use ark_relations::ns;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize, SerializationError};
 use core::marker::PhantomData;
-use derivative::Derivative;
 use manta_crypto::{
     constraint::{Allocator, Constant, Equal, Public, Secret, ValueSource, Variable},
-    ecc,
-    ecc::{PointAdd, PointDouble},
+    ecc::{self, PointAdd, PointDouble},
     key::kdf,
     rand::{CryptoRng, RngCore, Sample, Standard},
 };
@@ -368,7 +366,7 @@ where
 }
 
 /// Elliptic Curve Group Element Variable
-#[derive(Derivative)]
+#[derive(derivative::Derivative)]
 #[derivative(Clone)]
 pub struct GroupVar<C, CV>(pub(crate) CV, PhantomData<C>)
 where
@@ -573,7 +571,7 @@ where
 }
 
 #[cfg(test)]
-mod tests {
+mod test {
     use super::*;
     use ark_ec::ProjectiveCurve;
     use ark_ed_on_bls12_381::{constraints::EdwardsVar, EdwardsProjective};
@@ -599,20 +597,14 @@ mod tests {
         const NUM_TRIALS: usize = 5;
 
         let mut cs = R1CS::for_known();
-
         for _ in 0..NUM_TRIALS {
             let base = Group::<_>::gen(rng).as_known::<Secret, GroupVar<_, _>>(&mut cs);
-
             let scalar = Scalar::<C>::gen(rng).as_known::<Secret, _>(&mut cs);
-
-            let expected = ScalarMul::scalar_mul(&base, &scalar, &mut cs);
-
+            let expected = base.scalar_mul(&scalar, &mut cs);
             let table = PreprocessedScalarMulTable::<_, N>::from_base(base, &mut cs);
             let actual = table.scalar_mul(&scalar, &mut cs);
-
             cs.assert_eq(&expected, &actual);
         }
-
         assert!(cs.is_satisfied());
     }
 
--- a/manta-pay/src/crypto/ecc/arkworks.rs
+++ b/manta-pay/src/crypto/ecc/arkworks.rs
@@ -35,6 +35,7 @@ use manta_util::codec;
 use manta_util::serde::{Deserialize, Serialize, Serializer};
 
 pub use ark_ec::{AffineCurve, ProjectiveCurve};
+use ark_ed_on_bls12_381::EdwardsProjective;
 pub use ark_r1cs_std::groups::CurveVar;
 
 /// Constraint Field Type
@@ -406,30 +407,42 @@ where
     }
 }
 
-impl<C, CV, const N: usize> ecc::PreprocessedScalarMul<Compiler<C>, N> for GroupVar<C, CV> where
-    C: ProjectiveCurve,
-    CV: CurveVar<C, ConstraintField<C>> {
-    /// Performs the scalar multiplication against a pre-computed table.
-    ///
-    /// The pre-computed table is a list of power-of-two multiples of `scalar`, such that
-    /// `table[i] = scalar * 2^i`.
-    fn preprocessed_scalar_mul(table: &[Self; N], scalar: &Self::Scalar, compiler: &mut Compiler<C>) -> Self::Output {
-        // Computes the standard little-endian double-and-add algorithm
-        // (Algorithm 3.26, Guide to Elliptic Curve Cryptography)
-        // adapted from https://github.com/arkworks-rs/r1cs-std/blob/50ab8ee5ba8c09637044718eceac38e98e0ea67c/src/groups/mod.rs#L110-L139
+macro_rules! impl_processed_scalar_mul {
+    ($curve: ty, $N: expr) => {
+        impl<CV> ecc::PreprocessedScalarMul<Compiler<$curve>, $N> for GroupVar<$curve, CV>
+        where
+            CV: CurveVar<$curve, ConstraintField<EdwardsProjective>>,
+        {
+            fn preprocessed_scalar_mul(
+                table: &[Self; $N],
+                scalar: &Self::Scalar,
+                compiler: &mut Compiler<EdwardsProjective>,
+            ) -> Self::Output {
+                // Computes the standard little-endian double-and-add algorithm
+                // (Algorithm 3.26, Guide to Elliptic Curve Cryptography)
+                // adapted from https://github.com/arkworks-rs/r1cs-std/blob/50ab8ee5ba8c09637044718eceac38e98e0ea67c/src/groups/mod.rs#L110-L139
 
-        let _ = compiler;
-        let mut result: CV = CV::zero();
-        let scalar_bits = scalar.0.to_bits_le().expect("Bit decomposition is not allowed to fail.");
-        debug_assert_eq!(scalar_bits.len(), N, "Scalar is expected to have N bits.");
-        for (bit, base) in scalar_bits.into_iter().zip(table.iter()){
-            // compute `self + 2^i * scalar`
-            let scalar_plus_base = result.clone() + base.0.clone();
-            result = bit.select(&scalar_plus_base, &result).expect("Conditional select is not allowed to fail. ");
+                let _ = compiler;
+                let mut result: CV = CV::zero();
+                let scalar_bits = scalar
+                    .0
+                    .to_bits_le()
+                    .expect("Bit decomposition is not allowed to fail.");
+                debug_assert_eq!(scalar_bits.len(), $N, "Scalar is expected to have N bits.");
+                for (bit, base) in scalar_bits.into_iter().zip(table.iter()) {
+                    // compute `self + 2^i * scalar`
+                    let scalar_plus_base = result.clone() + base.0.clone();
+                    result = bit
+                        .select(&scalar_plus_base, &result)
+                        .expect("Conditional select is not allowed to fail. ");
+                }
+                Self(result, PhantomData)
+            }
         }
-        Self(result, PhantomData)
-    }
+    };
 }
+
+impl_processed_scalar_mul!(EdwardsProjective, 252);
 
 impl<C, CV> Equal<Compiler<C>> for GroupVar<C, CV>
 where
--- a/manta-pay/src/crypto/ecc/arkworks.rs
+++ b/manta-pay/src/crypto/ecc/arkworks.rs
@@ -572,11 +572,12 @@ where
 mod tests {
     use super::*;
     use ark_ec::ProjectiveCurve;
-    use ark_ed_on_bls12_381::constraints::EdwardsVar;
-    use ark_ed_on_bls12_381::EdwardsProjective;
-    use manta_crypto::constraint::ConstraintSystem;
-    use manta_crypto::ecc::{PreprocessedScalarMulTable, ScalarMul};
-    use manta_crypto::rand::OsRng;
+    use ark_ed_on_bls12_381::{constraints::EdwardsVar, EdwardsProjective};
+    use manta_crypto::{
+        constraint::ConstraintSystem,
+        ecc::{PreprocessedScalarMulTable, ScalarMul},
+        rand::OsRng,
+    };
     fn preprocessed_scalar_mul_test_template<C, CV, const N: usize>(rng: &mut OsRng)
     where
         C: ProjectiveCurve,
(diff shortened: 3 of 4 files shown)


manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +406,31 @@ where
...
+        let scalar_bits = scalar.0.to_bits_le().expect("Bit decomposition is not allowed to fail.");
+        debug_assert_eq!(scalar_bits.len(), N, "Scalar is expected to have N bits.");
+        for (bit, base) in scalar_bits.into_iter().zip(table.iter()){
+            // compute `self + 2^i * scalar`
+            let scalar_plus_base = result.clone() + base.0.clone();
+            result = bit.select(&scalar_plus_base, &result).expect("Conditional select is not allowed to fail. ");
+        }
+        Self(result, PhantomData)
+    }
+}
```
bhgomes, line 432: The `N` here represents the compile-time fixed and known number of bits of a scalar for the particular curve. This trait should not be implemented over all `N` (I wanted to use an associated constant but it's unstable to use that with slices rn). We can improve the documentation to help make this clearer.

manta-crypto/src/ecc.rs
```diff
@@ -145,20 +145,6 @@ pub trait PreprocessedScalarMul<COM, const N: usize>: ScalarMul<COM> + Sized {
...
-{
-    #[inline]
-    fn preprocessed_scalar_mul(
-        table: &[Self; 1],
-        scalar: &Self::Scalar,
-        compiler: &mut COM,
-    ) -> Self::Output {
-        table[0].scalar_mul(scalar, compiler)
-    }
-}
```
bhgomes, line 160: Because the trait should be implemented for a single value of `N`, this should not be deleted.

bhgomes (changes requested):

manta-crypto/src/ecc.rs
tsunrise: Implementation of `1` conflicts with implementation of `N`. Any good way to bypass this?

tsunrise: ah ok, I see the previous comment. so we are not implementing this for all `N`?

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: To which specific value of `N` shall we implement? I can write a macro to do that

bhgomes: It should be implemented for the number of bits in a scalar. It's exactly the "debug assert" that you do, but we know at compile-time that the table is the correct size.

manta-crypto/src/ecc.rs
bhgomes: Yes, only for the exact `N` which is correct.

bhgomes (changes requested):

manta-crypto/src/ecc.rs
tsunrise: Got it thanks!

tsunrise: Side note: currently we cannot put associate constant to const generic. For example, we cannot implement `ecc::PreprocessedScalarMul<Compiler<C>, Self::Scalar::MODULUS_BITS>`. We might need to implement that for every individual field.

tsunrise: I have implemented this trait only for Bls12_381. I also wrote a macro to easily support other curves if we need that

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: currently we cannot put associate constant to const generic. For example, we cannot implement `ecc::PreprocessedScalarMul<Compiler<C>, Self::Scalar::MODULUS_BITS>`. We might need to implement that for every individual curve. 

I have implemented this trait only for Bls12_381. I also wrote a macro to easily support other curves if we need that

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +406,31 @@ where
     }
 }
 
+impl<C, CV, const N: usize> ecc::PreprocessedScalarMul<Compiler<C>, N> for GroupVar<C, CV> where
+    C: ProjectiveCurve,
+    CV: CurveVar<C, ConstraintField<C>> {
+    /// Performs the scalar multiplication against a pre-computed table.
+    ///
+    /// The pre-computed table is a list of power-of-two multiples of `scalar`, such that
+    /// `table[i] = scalar * 2^i`.
```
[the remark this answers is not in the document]
tsunrise: fixed

tsunrise:
Can someone allow force push on branches other than `main`? I messed up with commit histories @bhgomes @stechu

tsunrise:
Can someone allow force push on branches other than `main`? I messed up with commit histories @bhgomes @stechu

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +407,43 @@ where
...
+                        .select(&scalar_plus_base, &result)
+                        .expect("Conditional select is not allowed to fail. ");
+                }
+                Self(result, PhantomData)
+            }
+        }
+    };
+}
+
+impl_processed_scalar_mul!(EdwardsProjective, 252);
```
bhgomes, line 445: You can use the compile-time known size:

```rust
impl_processed_scalar_mul!(EdwardsProjective, {
    <<EdwardsProjective as ProjectiveCurve>::ScalarField as PrimeField>::Params::MODULUS_BITS
        as usize
});
```

We can incorporate this into the macro eventually to make it easier to implement.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +407,43 @@ where
     }
 }
 
+macro_rules! impl_processed_scalar_mul {
+    ($curve: ty, $N: expr) => {
+        impl<CV> ecc::PreprocessedScalarMul<Compiler<$curve>, $N> for GroupVar<$curve, CV>
+        where
+            CV: CurveVar<$curve, ConstraintField<EdwardsProjective>>,
+        {
+            fn preprocessed_scalar_mul(
```
bhgomes, line 416: Please `#[inline]` this method.

bhgomes (changes requested):
Left some comments on the implementation that should be addressed before merging. I've also left some notes on some infrastructure we should add to improve development in ECC.

manta-crypto/src/ecc.rs
```diff
@@ -137,6 +137,9 @@ pub trait ScalarMul<COM = ()> {
 /// Elliptic Curve Pre-processed Scalar Multiplication Operation
 pub trait PreprocessedScalarMul<COM, const N: usize>: ScalarMul<COM> + Sized {
     /// Performs the scalar multiplication against a pre-computed table.
+    ///
+    /// The pre-computed table is a list of power-of-two multiples of `scalar`, such that
+    /// `table[i] = scalar * 2^i`.
```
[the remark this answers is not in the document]
tsunrise: resolved offline

manta-crypto/src/ecc.rs
```diff
@@ -137,6 +137,9 @@ pub trait ScalarMul<COM = ()> {
 /// Elliptic Curve Pre-processed Scalar Multiplication Operation
 pub trait PreprocessedScalarMul<COM, const N: usize>: ScalarMul<COM> + Sized {
```
tsunrise, line 138: I will add some tests for this in a day or two. Do not merge until this comment is marked resolved

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +407,43 @@ where
...
+                table: &[Self; $N],
+                scalar: &Self::Scalar,
+                compiler: &mut Compiler<EdwardsProjective>,
+            ) -> Self::Output {
+                // Computes the standard little-endian double-and-add algorithm
+                // (Algorithm 3.26, Guide to Elliptic Curve Cryptography)
+                // adapted from https://github.com/arkworks-rs/r1cs-std/blob/50ab8ee5ba8c09637044718eceac38e98e0ea67c/src/groups/mod.rs#L110-L139
+
```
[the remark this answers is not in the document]
tsunrise: I agree. Removed.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +407,43 @@ where
...
+                    .0
+                    .to_bits_le()
+                    .expect("Bit decomposition is not allowed to fail.");
+                debug_assert_eq!(scalar_bits.len(), $N, "Scalar is expected to have N bits.");
+                for (bit, base) in scalar_bits.into_iter().zip(table.iter()) {
+                    // compute `self + 2^i * scalar`
+                    let scalar_plus_base = result.clone() + base.0.clone();
+                    result = bit
+                        .select(&scalar_plus_base, &result)
+                        .expect("Conditional select is not allowed to fail. ");
```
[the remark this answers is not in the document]
tsunrise: resolve first comment.  For second comment, see #39

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +407,43 @@ where
...
+
+                let _ = compiler;
+                let mut result: CV = CV::zero();
+                let scalar_bits = scalar
+                    .0
+                    .to_bits_le()
+                    .expect("Bit decomposition is not allowed to fail.");
+                debug_assert_eq!(scalar_bits.len(), $N, "Scalar is expected to have N bits.");
+                for (bit, base) in scalar_bits.into_iter().zip(table.iter()) {
+                    // compute `self + 2^i * scalar`
```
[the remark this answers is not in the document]
tsunrise: removed

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +407,43 @@ where
...
+macro_rules! impl_processed_scalar_mul {
+    ($curve: ty, $N: expr) => {
+        impl<CV> ecc::PreprocessedScalarMul<Compiler<$curve>, $N> for GroupVar<$curve, CV>
+        where
+            CV: CurveVar<$curve, ConstraintField<EdwardsProjective>>,
+        {
+            fn preprocessed_scalar_mul(
+                table: &[Self; $N],
+                scalar: &Self::Scalar,
+                compiler: &mut Compiler<EdwardsProjective>,
```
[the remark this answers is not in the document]
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -406,6 +407,43 @@ where
...
+            ) -> Self::Output {
+                // Computes the standard little-endian double-and-add algorithm
+                // (Algorithm 3.26, Guide to Elliptic Curve Cryptography)
+                // adapted from https://github.com/arkworks-rs/r1cs-std/blob/50ab8ee5ba8c09637044718eceac38e98e0ea67c/src/groups/mod.rs#L110-L139
+
+                let _ = compiler;
+                let mut result: CV = CV::zero();
```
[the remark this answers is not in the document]
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-crypto/src/ecc.rs
tsunrise: test added

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -36,6 +36,7 @@ use manta_util::serde::{Deserialize, Serialize, Serializer};
 
 pub use ark_ec::{AffineCurve, ProjectiveCurve};
 pub use ark_r1cs_std::groups::CurveVar;
+use manta_crypto::ecc::{PointAdd, PointDouble};
```
bhgomes, line 39: Move this non-public `use` import together with the other ones. The ordering of imports is

1. `use`
2. `pub use`
3. `mod`
4. `pub mod`

So non-`pub` before `pub`. Also, non-`feature` before `feature` gated and features in alphabetical order.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -363,6 +364,7 @@ where
 }
 
 /// Elliptic Curve Group Element Variable
+#[derive(Clone)]
```
bhgomes, line 367: Please use `derivative::Derivative` for `derive` traits when there are type parameters on the deriving type so we have the correct semantics.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+impl<C, CV> PointAdd<Compiler<C>> for GroupVar<C, CV>
+where
+    C: ProjectiveCurve,
+    CV: CurveVar<C, ConstraintField<C>>,
+{
+    type Output = Self;
+
+    fn add(&self, rhs: &Self, compiler: &mut Compiler<C>) -> Self::Output {
+        let _ = compiler;
+        let mut result = self.0.clone();
```
bhgomes, line 552: Do we really need this clone? Can we not do `Self::new(&self.0 + &rhs.0)`? I believe `+` already does `+=` internally so it won't have a performance penalty.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ark_ec::ProjectiveCurve;
+    use ark_ed_on_bls12_381::constraints::EdwardsVar;
+    use ark_ed_on_bls12_381::EdwardsProjective;
+    use manta_crypto::constraint::ConstraintSystem;
+    use manta_crypto::ecc::{PreprocessedScalarMulTable, ScalarMul};
+    use manta_crypto::rand::OsRng;
+    fn preprocessed_scalar_mul_test_template<C, CV, const N: usize>(rng: &mut OsRng)
```
bhgomes, line 582: Please leave a newline between the imports and the first definition. Also, add a documentation string here.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+            cs.assert_eq(&expected, &actual);
+        }
+
+        assert!(cs.cs.is_satisfied().unwrap());
+    }
+
+    #[test]
+    fn preprocessed_scalar_mul_test() {
+        let mut rng = OsRng::default();
+        preprocessed_scalar_mul_test_template::<EdwardsProjective, EdwardsVar, 252>(&mut rng);
```
bhgomes: You can just do something like:

```rust
preprocessed_scalar_mul_test_template::<EdwardsProjective, EdwardsVar, { 
    ScalarParam<EdwardsProjective>::MODULUS_BITS
}>(&mut OsRng);
```

however it gets formatted, let `rustfmt` decide.

manta-pay/Cargo.toml
```diff
@@ -78,7 +78,7 @@ simulation = [
 std = ["manta-accounting/std", "manta-util/std"]
 
 # Testing Frameworks
-test = ["manta-accounting/test", "manta-crypto/test"]
+test = ["manta-accounting/test", "manta-crypto/test", "manta-crypto/getrandom"]
```
bhgomes, line 81: Instead of adding to the `test` feature, you should add this to the `dev-dependencies`. The `test` feature is for _exported_ test content rather than non-exported tests. NB: the `#[cfg(test)]` does not observe this feature only `#[cfg(feature = "test")]` does.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -408,49 +410,41 @@ where
...
-impl<CV>
-    ecc::PreprocessedScalarMul<
-        Compiler<$curve>,
+        impl<CV>
+            ecc::PreprocessedScalarMul<
+                Compiler<$curve>,
+                {
+                    <<$curve as ProjectiveCurve>::ScalarField as PrimeField>::Params::MODULUS_BITS
+                        as usize
+                },
```
bhgomes, line 419: Can you try to make an alias for this? Not sure if it'll work because it may involve an unstable feature, but worth a try:

```rust
type ScalarParams<C> = <<C as ProjectiveCurve>::ScalarField as PrimeField>::Params;
```

Then you can call it as `{ ScalarParams::<$curve>::MODULUS_BITS }`. Also, we can call this from the test.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
 }
+
+impl<C, CV> PointAdd<Compiler<C>> for GroupVar<C, CV>
+where
+    C: ProjectiveCurve,
+    CV: CurveVar<C, ConstraintField<C>>,
+{
+    type Output = Self;
+
+    fn add(&self, rhs: &Self, compiler: &mut Compiler<C>) -> Self::Output {
```
bhgomes, line 550: Please `#[inline]`. I'd like to inline all `impl Trait` methods for now.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+}
+
+impl<C, CV> PointDouble<Compiler<C>> for GroupVar<C, CV>
+where
+    C: ProjectiveCurve,
+    CV: CurveVar<C, ConstraintField<C>>,
+{
+    type Output = Self;
+
+    fn double(&self, compiler: &mut Compiler<C>) -> Self::Output {
```
bhgomes, line 565: Please `#[inline]`. I'd like to inline all `impl Trait` methods for now.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ark_ec::ProjectiveCurve;
+    use ark_ed_on_bls12_381::constraints::EdwardsVar;
+    use ark_ed_on_bls12_381::EdwardsProjective;
+    use manta_crypto::constraint::ConstraintSystem;
+    use manta_crypto::ecc::{PreprocessedScalarMulTable, ScalarMul};
+    use manta_crypto::rand::OsRng;
```
bhgomes: Make sure to run `cargo fmt` on nightly to use the formating conventions the CI will accept.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+    fn preprocessed_scalar_mul_test_template<C, CV, const N: usize>(rng: &mut OsRng)
+    where
+        C: ProjectiveCurve,
+        CV: CurveVar<C, ConstraintField<C>>,
+        GroupVar<C, CV>: ecc::PreprocessedScalarMul<
```
bhgomes, line 586: Feel free to import `PreprocessedScalarMul` to avoid `ecc::` prefix. It's more important the local code is as simple as possible even if we have longer imports.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+
+            let scalar = Scalar::<C>::gen(rng);
+            let scalar_var = scalar.as_known::<Secret, ScalarVar<C, CV>>(&mut cs);
+
+            let expected = ScalarMul::scalar_mul(&base_var, &scalar_var, &mut cs);
+
+            let table = PreprocessedScalarMulTable::<_, N>::from_base(base_var, &mut cs);
+            let actual = table.scalar_mul(&scalar_var, &mut cs);
+
+            cs.assert_eq(&expected, &actual);
```
bhgomes, line 609: Feel free to use allocation strategies that reduce the amount of explicit type-annotations and trait names. Also, try to maximize the number of automatically deduced types by removing them one-by-one until either no types are specified or the compiler reaches an ambiguous typing state.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+        }
+
+        assert!(cs.cs.is_satisfied().unwrap());
```
bhgomes, line 612: If you'd like, add an `is_satisfied` method to the `cs` directly (the `arkworks` concrete type not the trait definition).

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+        const NUM_TRIALS: usize = 5;
+
+        let mut cs = R1CS::for_known();
+
+        for _ in 0..NUM_TRIALS {
+            let base = Group::<C>::gen(rng);
+            let base_var = <GroupVar<C, CV> as Variable<Secret, _>>::new_known(&base, &mut cs);
+
+            let scalar = Scalar::<C>::gen(rng);
+            let scalar_var = scalar.as_known::<Secret, ScalarVar<C, CV>>(&mut cs);
```
bhgomes, line 602: Unless you are also going to check something with the `base` and `scalar` directly, compress these groupings into

```rust
let base = Group::<C>::gen(rng).as_known::<Secret, _>(&mut cs);
```

(or whatever the minimally type-annotated form is) and likewise for `scalar`.

bhgomes (changes requested):
Just a few more changes and we'll be good to go!

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -545,3 +539,80 @@ where
...
+            let actual = table.scalar_mul(&scalar_var, &mut cs);
+
+            cs.assert_eq(&expected, &actual);
+        }
+
+        assert!(cs.cs.is_satisfied().unwrap());
+    }
+
+    #[test]
+    fn preprocessed_scalar_mul_test() {
```
bhgomes, line 616: You should leave off `test` prefixes and suffixes from the names of tests since `mod test` will already pick up a `test::` prefix.

bhgomes (changes requested):
One more thing about test naming.

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: unfortunately `CurveVar` does not implement `Add` trait :( We can upstream this impl to arkworks in the future

manta-pay/Cargo.toml
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: it works! done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -23,9 +23,11 @@ use ark_r1cs_std::ToBitsGadget;
 use ark_relations::ns;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize, SerializationError};
 use core::marker::PhantomData;
+use derivative::Derivative;
```
bhgomes, line 26: I'm using the convention to inline this macro at the call site to reduce the number of imports, since this macro is used a lot, and importing this does not reduce the complexity at the call site:

```rust
```

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -23,9 +23,11 @@ use ark_r1cs_std::ToBitsGadget;
...
 use manta_crypto::{
     constraint::{Allocator, Constant, Equal, Public, Secret, ValueSource, Variable},
     ecc,
+    ecc::{PointAdd, PointDouble},
```
bhgomes, line 30: For these imports, use `self`:

```rust
ecc::{self, PointAdd, PointDouble}
```

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -495,3 +541,88 @@ where
...
+
+    #[inline]
+    fn double(&self, compiler: &mut Compiler<C>) -> Self::Output {
+        let _ = compiler;
+        Self::new(self.0.double().expect("Doubling is not allowed to fail."))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
bhgomes, line 576: We use the name `test` for this module by convention. Unless specifically done so, all modules should be named in singular form rather than plural.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -495,3 +541,88 @@ where
...
+
+            let expected = ScalarMul::scalar_mul(&base, &scalar, &mut cs);
+
+            let table = PreprocessedScalarMulTable::<_, N>::from_base(base, &mut cs);
+            let actual = table.scalar_mul(&scalar, &mut cs);
+
+            cs.assert_eq(&expected, &actual);
+        }
+
+        assert!(cs.is_satisfied());
```
bhgomes, line 616: Please remove excess whitespace and use `base.scalar_mul(&scalar, &mut cs)` for `expected`.

bhgomes (changes requested):

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-pay/src/crypto/ecc/arkworks.rs
tsunrise: done

manta-crypto/src/ecc.rs
```diff
@@ -162,7 +165,7 @@ where
 /// Elliptic Curve Group
 pub trait Group<COM = ()>: PointAdd<COM> + PointDouble<COM> + ScalarMul<COM> {}
 
-/// Pre-processed Scalar Multiplication Table
+/// Pre-processed Scalar Multiplication Table, represented as powers of two of `scalar`.
```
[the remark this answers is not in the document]
tsunrise: done

bhgomes (approved):

PR #36, opened 2022-03-10
Simplify errors

stechu:
Get rid of stand alone `LedgerInternalError`

### DIFF
--- a/manta-accounting/src/transfer/mod.rs
+++ b/manta-accounting/src/transfer/mod.rs
@@ -2037,8 +2037,8 @@ pub struct InvalidSourceAccount<AccountId> {
     /// Account Id
     pub account_id: AccountId,
 
-     /// Asset Id
-     pub asset_id: AssetId,
+    /// Asset Id
+    pub asset_id: AssetId,
 
     /// Amount Attempting to Withdraw
     pub withdraw: AssetValue,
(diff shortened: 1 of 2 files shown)


bhgomes (changes requested):
What I meant for structuring the errors was something like the following:

```rust
trait TransferLedger {
    /// State Update Error
    type UpdateError;

    fn update_public_balances(...) -> Result<(), Self::UpdateError>;
}
```

where this error is `core::convert::Infallible` for the mock ledger and for substrate it can be some other error.

Also, please remove the dead code introduced by the previous PR, namely the `AccountBalance` enum and the `LedgerInternalError` struct.

manta-accounting/src/transfer/mod.rs
```diff
@@ -1920,6 +1920,9 @@ where
     /// Ledger Event
     type Event;
 
+    /// State Update Error
```
bhgomes, line 1923: Can we add the following full comment:

```rust
/// State Update Error
///
/// This error type is used if the ledger can fail when updating the public state. The
/// [`update_public_balances`](Self::update_public_balances) method uses this error type to
/// track this condition.
type UpdateError;
```

with this exact formatting.

bhgomes (changes requested):
We basically want the ledger to state whether updating the public balances can fail or not. In the case of the mock ledger, it cannot so the error type is `Infallible` but for substrate, we may have some more complex requirements. Eventually, we want to revisit the error states of the ledgers since they assume no errors during other state updates but we may want to make this explicit. Currently, the API is a bit asymmetric because only `update_public_balances` gets an error condition, but this is fine for now.

bhgomes (approved):

Issue #22, opened 2022-02-27
Use Fastest Poseidon Implementation


bhgomes:
We have a few different Poseidon implementations in the Manta ecosystem and we want to use the fastest one in production. Here are some goals for merging the implementation from [Plonk-Prototype](https://github.com/manta-network/Plonk-Prototype):

- Upgrade `posiedon` file to directory in `manta-pay` crate
- Keep the same interface that is defined in the current `poseidon` module, just upgrade the implementation (should not need to change any other files in `manta-pay` after the upgrade)
- Add tests for consistency with known Poseidon implementations
- Add in the parameter generation logic into the `posiedon` module and upgrade the parameters in the `bin/generate_parameters.rs` file (also remove the `Sample` implementation since it is not correct)
- Add a PR to [manta-sdk](https://github.com/manta-network/sdk) for new parameters

NB: For this PR we only care about the R1CS implementation, not PLONK.

tsunrise:
I will start this soon

PR #51, opened 2022-04-08
Use Poseidon Implementation from `plonk-prototype`

tsunrise:
Use Poseidon Implementation from `plonk-prototype`. Compare to current Poseidon implementation, this PR will add parameter generation and filecoin optimization. 

🚧 **WIP**

closes: #22


tsunrise:
Something worth discussing at this point: 
- Are we going to calculate round constants at compile time or runtime? It might be tricky to calculate round constants in compile time, but we can always hardcode them. 
- If we hardcode those round constants, are we going to put it as an associated constants? We cannot do this if we want the poseidon state to be an array because `const_generics` is not supported yet.

bhgomes:
> Something worth discussing at this point: 
> 
> - Are we going to calculate round constants at compile time or runtime? It might be tricky to calculate round constants in compile time, but we can always hardcode them. 
> 
> - If we hardcode those round constants, are we going to put it as an associated constants? We cannot do this if we want the poseidon state to be an array because `const_generics` is not supported yet. 

We can calculate round constants at runtime, they are part of the "parameter sampling" pipeline. This is only ever run once anyway so it's not important to make it a const-expr.

tsunrise:
Changes in this PR is moved to #57

PR #57, opened 2022-05-02
Feature: Parameter Generation for Poseidon Hash

BoyuanFeng:
Integrate from Plonk Prototype to Manta-rs

Major changes:

- Extend the `Specification` trait in `manta-pay/src/crypto/hash/poseidon.rs` to 1) separate fields for parameters (public constants) and states (private variables); 2) generically support addition, multiplication, etc
- Rewrite parameter generation following the `Specification` trait such that the code can work for generic fields
- Update Poseidon hash implementation to match #constraints between manta-rs and Plonk Prototype. Saw 20~30% saving in #constraints and latency than previous manta-rs, in terms of Poseidon hash, mint, reclaim, and private transfer.

Changes in #51 are moved here.
Closes: #22


manta-pay/src/crypto/hash/poseidon_parameter_generation/constants.rs
```diff
@@ -0,0 +1,109 @@
...
+use super::round_numbers::calc_round_numbers;
+use super::round_constants::generate_round_constants;
+use super::preprocessing::compress_round_constants;
+use super::matrix::Matrix;
+use super::mds::{factor_to_sparse_matrixes, MdsMatrices, SparseMatrix};
+use crate::crypto::hash::poseidon::Specification;
+use core::fmt::Debug;
+use alloc::vec::Vec;
+
+// TODO: shall we put constant generation code to compile time?
```
tsunrise, line 28: Concluded in last meeting, we don't need to put constant generation in compile time (should be in manta-sdk?)

manta-pay/src/crypto/hash/poseidon.rs
```diff
@@ -81,30 +128,30 @@ where
...
     /// Additive Round Keys
-    additive_round_keys: Vec<S::Field>,
+    additive_round_keys: Vec<S::ParameterField>,
 
     /// MDS Matrix
-    mds_matrix: Vec<S::Field>,
+    mds_matrix: Vec<S::ParameterField>,
 }
 
 impl<S, COM, const ARITY: usize> Hasher<S, COM, ARITY>
```
tsunrise, line 157: I remember in original implementation by @bhgomes, the ARITY is associated constant? We might condier doing so as well?

manta-pay/src/crypto/hash/poseidon.rs
```diff
@@ -188,8 +235,8 @@ where
     #[inline]
     fn first_round(&self, input: [&S::Field; ARITY], compiler: &mut COM) -> State<S, COM> {
```
tsunrise, line 236: If `ARITY` is associated constant, we have use slice here. Also, `[&S::Field; ARITY]` does not make sense to me. It should either be `&[S::Field; ARITY]` (if we use ARITY as const generics), or `&[S::Field]` and check arity on runtime (if we use ARITY as associated type constant)

manta-pay/src/crypto/hash/poseidon_parameter_generation/constants.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// a vector of round constants
+    pub round_constants: Vec<S::ParameterField>,
+    /// The following three operators are used for optimized poseidon hash
+    pub compressed_round_constants: Vec<S::ParameterField>,
+    pub pre_sparse_matrix: Matrix<S, COM>,
+    pub sparse_matrixes: Vec<SparseMatrix<S, COM>>,
+    /// domain tag for domain separation
+    pub domain_tag: S::ParameterField,
+    /// number of full rounds. Note: full_rounds % 2 == 0
+    pub full_rounds: usize,
```
tsunrise, line 48: We might want to hardcode `full_rounds`, `half_full_rounds`, and `partial_rounds` as a spec (e.g. associated constant of a trait) instead of calculating on runtime. Check @bhgomes's original implementation for an example.

tsunrise: This requires some nontrivial work and careful design. We can do this later if you want, or I can handle this when I have time.

manta-pay/src/crypto/hash/poseidon_parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,210 @@
...
+    #[test]
+    fn test_grain_lfsr_consistency() {
+        // sage generate_parameters_grain_deterministic.sage 1 0 255 3 8 55 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001
+
+        let mut lfsr = GrainLFSR::new(255, 3, 8, 55);
+        assert_eq!(
+            lfsr.get_field_elements_rejection_sampling::<PoseidonSpec, Compiler<PoseidonSpec>>(1)[0],
+            Fp(field_new!(
+                ConstraintField,
+                "41764196652518280402801918994067134807238124178723763855975902025540297174931"
```
tsunrise, line 199: We can add the sage script in plonk-prototype to show how to generate the test case for this hardcode test. Those scripts should be correct (or I messed up)

tsunrise (commented):
Just took a very rough look at the code. Have a few comments. Will do it again in a day or two!

manta-pay/src/crypto/hash/poseidon.rs
bhgomes: The `[&S::Field; ARITY]` type means that we are constraining the input to be `ARITY`-many references rather than `&[S::Field; ARITY]` which is a reference to an array of values. The first one is more conservative because the values can come from anywhere, we just point to them, whereas the second one requires that the values are all packed contiguously and if they are not, it requires moving the values first before passing them into the function.

manta-pay/src/crypto/hash/poseidon_parameter_generation/constants.rs
BoyuanFeng: This part is only used in the optimized Poseidon hash. Considering 1) this PR is already kind of complex, and 2) the optimized Poseidon hash does not bring benefits on Groth16, I would like to suggest keeping this part and update in the next PR when we adapt the optimized Poseidon hash.

manta-pay/src/crypto/hash/poseidon.rs
BoyuanFeng: Do you mean in the `pub struct Hasher`? It seems there are only additive_round_keys and mds_matrix in the original implementation by @bhgomes 
See https://github.com/Manta-Network/manta-rs/blob/main/manta-pay/src/crypto/hash/poseidon.rs#L99-L108

manta-pay/src/crypto/hash/poseidon_parameter_generation/constants.rs
BoyuanFeng: @bhgomes what's your opinion on this issue?

bhgomes: No need, just try to put the spec generation at compile time like the number of full and partial rounds and the alpha coefficient.

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+
+use crate::crypto::hash::ParamField;
+
+pub struct GrainLFSR {
+    state: [bool; 80],
+    prime_num_bits: u64,
+    head: usize,
+}
+
+fn append_bits<T: Into<u128>>(state: &mut [bool; 80], head: &mut usize, n: usize, from: T) {
```
bhgomes, line 30:
```suggestion
fn append_bits<T>(state: &mut [bool; 80], head: &mut usize, n: usize, from: T) 
where
    T: Into<u128>,
{

```

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+    let val = from.into() as u128;
+    for i in (0..n).rev() {
+        state[*head] = (val >> i) & 1 != 0;
+        *head += 1;
+        *head %= 80;
+    }
+}
+
+impl GrainLFSR {
+    pub fn new(prime_num_bits: u64, width: usize, r_f: usize, r_p: usize) -> Self {
```
bhgomes, line 40: Add docs and use full names for these inputs.

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+        // b40..=49 describes r_p (num_partial_rounds)
+        append_bits(&mut init_sequence, &mut head, 10, r_p as u16);
+        // b50..=79 describes the constant 1
+        append_bits(
+            &mut init_sequence,
+            &mut head,
+            30,
+            0b111111111111111111111111111111u128,
+        );
+        let mut res = GrainLFSR {
```
bhgomes, line 62:
```suggestion
        let mut res = Self {

```

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+            state: init_sequence,
+            prime_num_bits,
+            head,
+        };
+        res.init();
+        res
+    }
+
+    fn update(&mut self) -> bool {
```
bhgomes, line 71: Add docs

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+
+            res.push(F::from_le_bytes_mod_order(&bytes));
+        }
+
+        res
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
bhgomes, line 169: Use `test` singular form

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+                    res.push(f);
+                    break;
+                }
+            }
+        }
+
+        res
+    }
+
+    pub fn get_field_elements_mod_p<F>(&mut self, num_elems: usize) -> Vec<F>
```
bhgomes, line 138: Add docs and change argument to `count`

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+                        result |= u8::from(*bit) << i
+                    }
+                    result
+                })
+                .collect::<Vec<u8>>();
+
+            res.push(F::from_le_bytes_mod_order(&bytes));
+        }
+
+        res
```
bhgomes, line 164:
```suggestion
        assert_eq!(F::MODULUS_BITS as u64, self.prime_num_bits);
        let mut field_elements = Vec::new();
        for _ in 0..num_elems {
            let mut bits = self.get_bits(self.prime_num_bits as usize);
            bits.reverse();
            let bytes = bits
                .chunks(8)
                .map(|chunk| {
                    let mut result = 0u8;
                    for (i, bit) in chunk.iter().enumerate() {
                        result |= u8::from(*bit) << i
                    }
                    result
                })
                .collect::<Vec<u8>>();
            field_elements.push(F::from_le_bytes_mod_order(&bytes));
        }
        field_elements
```

bhgomes: Separate 

```rust
let mut result = 0u8;
for (i, bit) in chunk.iter().enumerate() {
    result |= u8::from(*bit) << i
}
result
```

into its own function. Also, remember to use relevant names where possible: `result` doesn't describe what it is.

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+                // Constructs the number
+                if let Some(f) = F::try_from_bits_le(&bits) {
+                    res.push(f);
+                    break;
+                }
+            }
+        }
+
+        res
+    }
```
bhgomes, line 136: Same comments as below and remove internal function docs that explain "what" is happening. Instead always try to answer a "why" question when adding internal function `NOTE` comments.

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+            self.update();
+        }
+    }
+
+    #[inline]
+    fn bit(&self, index: usize) -> bool {
+        self.state[(index + self.head) % 80]
+    }
+
+    pub fn get_bits(&mut self, num_bits: usize) -> Vec<bool> {
```
bhgomes, line 91: Add docs

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+        self.head += 1;
+        self.head %= 80;
+        new_bit
+    }
+
+    fn init(&mut self) {
+        for _ in 0..160 {
+            self.update();
+        }
+    }
```
bhgomes, line 84: This function is only called in the constructor. It's ok to separate it to manage complexity, but since you have control over the interface, design it to reduce code in the constructor. For example:

```suggestion
    /// ADD DOCS HERE
    fn init(mut self) -> Self {
        for _ in 0..160 {
            self.update();
        }
        self
    }
```

Also, remove any magic constants like `160` and move them to `struct` `const`s.

bhgomes (changes requested):
Small review right now, more comments later, but please address the current comments first. Thanks!

manta-pay/src/crypto/hash/poseidon.rs
tsunrise: yes

manta-pay/src/crypto/hash/poseidon.rs
tsunrise: oh that makes sense, you are right

manta-pay/src/crypto/hash/poseidon_parameter_generation/lfsr.rs
tsunrise: I created an issue #67 for this

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
tsunrise:
```suggestion
/// Return a new `GrainLFSR` for poseidon parameter generation.
pub fn new(prime_num_bits: u64, width: usize, full_rounds: usize, partial_rounds: usize) -> Self {   
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
tsunrise: Makes sense. I suggest adding constant `LFSR_SIZE`, and 160 will become `LFSR_SIZE * 2`

tsunrise (commented):
reviewed `hasher` and part of `lfsr`, will take a look at remaining part later

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -47,18 +51,28 @@ pub trait Specification<COM = ()> {
...
-    fn zero(compiler: &mut COM) -> Self::Field;
+    /// Returns the domain tag. We use different domain_tags different applications so that defending
+    /// against rainbow table attack.
+    fn domain_tag(compiler: &mut COM, arity: usize) -> Self::Field;
 
     /// Adds two field elements together.
     fn add(lhs: &Self::Field, rhs: &Self::Field, compiler: &mut COM) -> Self::Field;
 
+    /// Adds a field element with a constant
+    fn addi(lhs: &Self::Field, rhs: &Self::ParameterField, compiler: &mut COM) -> Self::Field;
```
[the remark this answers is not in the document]
BoyuanFeng: I have the same feeling as Tom. See https://www3.ntu.edu.sg/home/smitha/fyp_gerald/addiInstruction.html#:~:text=The%20ADDI%20instruction%20performs%20an,source%20register's%20address%2C%20immediate%20data.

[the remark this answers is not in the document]
bhgomes: The name `addi` is short-hand for "add immediate data". We should be using full names for variables and function names and `add_const` means "add constant", which in this case is clearer as you don't have to know what "immediate data means". I would even prefer `add_constant` but I won't go that explicit since `const` is already conventional language in Rust. Also, we should not be using register metaphors for these variables, they are more algebraic. If you look at how `Self::ParameterField` is actually allocated it uses the `Constant` trait, not the "immediate data" trait, so this matches closer to its meaning and naming conventions in the constraint system language.

Even if short-hands are conventional in some scenarios, unless matching exact compatibility with those systems (like building a register machine), we should prefer to be more explicit and assume no prior knowledge rather than being implicit and assuming familiarity with systems _other than_ the ones we are building.

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -355,35 +381,106 @@ pub mod arkworks {
...
+        fn domain_tag(_: &mut (), arity: usize) -> Self::Field {
+            Fp(S::Field::from(((1 << arity) - 1) as u64))
         }
 
         #[inline]
         fn add(lhs: &Self::Field, rhs: &Self::Field, _: &mut ()) -> Self::Field {
             Fp(lhs.0 + rhs.0)
         }
 
+        // When COM = (), we do not distinguish Field and ParameterField. So addi() has the same computation as add()
```
bhgomes, line 458:
```suggestion
```

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -230,56 +253,57 @@ where
     #[inline]
     fn hash_in(&self, input: [&Self::Input; ARITY], compiler: &mut COM) -> Self::Output {
         let mut state = self.first_round(input, compiler);
-        for round in 1..S::FULL_ROUNDS {
+        let half_full_round = S::FULL_ROUNDS / 2;
```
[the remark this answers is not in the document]
BoyuanFeng: it might be better to put in `Specification` trait? We can add `const HALF_FULL_ROUNDS` at the same position as `const FULL_ROUNDS`

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,806 @@
...
+/// Checks whether `elem` equals zero
+pub fn equal_zero<F>(elem: &F) -> bool
+where
+    F: ParamField,
+{
+    let zero = F::zero();
+    F::eq(elem, &zero)
+}
+
+#[cfg(test)]
```
tsunrise, line 497: add feature gate for `arkworks`

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,806 @@
...
+            vec![four, five, six],
+            vec![seven, eight, nine],
+        ]
+        .into();
+
+        let cases = [
+            (
+                0,
+                0,
+                Matrix::<Fp<Fr>>(vec![vec![five, six], vec![eight, nine]]),
```
tsunrise, line 526: I think type `Fr` can be inferred by compiler.replace it with `Fp<_>`?

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,806 @@
...
+    #[test]
+    fn test_scalar_mul() {
+        let zero = Fp(Fr::from(0u64));
+        let one = Fp(Fr::from(1u64));
+        let two = Fp(Fr::from(2u64));
+        let three = Fp(Fr::from(3u64));
+        let four = Fp(Fr::from(4u64));
+        let six = Fp(Fr::from(6u64));
+
+        let m = Matrix::<Fp<Fr>>(vec![vec![zero, one], vec![two, three]]);
```
tsunrise, line 585: `Fr` -> `_`

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,371 @@
...
+        MdsMatrices {
+            m: self.m.clone(),
+            m_inv: self.m_inv.clone(),
+            m_hat: self.m_hat.clone(),
+            m_hat_inv: self.m_hat_inv.clone(),
+            m_prime: self.m_prime.clone(),
+            m_double_prime: self.m_double_prime.clone(),
+        }
+    }
+}
```
tsunrise, line 58: derive `Clone` trait instead of explicitly implementing it

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,371 @@
...
+//! MDS Data Generation
+
+use crate::crypto::hash::{poseidon::parameter_generation::matrix::Matrix, ParamField};
+use alloc::{vec, vec::Vec};
+use core::fmt::Debug;
+
+/// MDS Matrix for both naive poseidon hash and optimized poseidon hash
+/// For detailed descriptions, please refer to <https://hackmd.io/8MdoHwoKTPmQfZyIKEYWXQ>
+/// Note: Naive and optimized poseidon hash does not change #constraints in Groth16.
+pub struct MdsMatrices<F>
```
tsunrise, line 26:
```suggestion
#[derive(Clone)]
pub struct MdsMatrices<F>
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,371 @@
...
+    sparse_matrices.reverse();
+    let sparse_matrixes = sparse_matrices
+        .iter()
+        .map(|m| SparseMatrix::<F>::new(m))
+        .collect::<Vec<_>>();
+
+    (pre_sparse, sparse_matrixes)
+}
+
+#[cfg(test)]
```
tsunrise, line 239: add feature gates for `arkworks`

manta-pay/src/crypto/hash/poseidon/parameter_generation/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Poseidon hash parameter generation
+
+mod lfsr;
+pub mod matrix;
+pub mod mds;
+pub mod preprocessing;
+pub mod round_constants;
```
tsunrise, line 23: Do we really want to expose all of them?

manta-pay/src/crypto/hash/poseidon/parameter_generation/preprocessing.rs
```diff
@@ -0,0 +1,100 @@
...
+    // Post S-box adds for the first set of full rounds should be 'inverted' from next round.
+    // The final round is skipped when fully preprocessing because that value must be obtained from the result of preprocessing the partial rounds.
+    let end = half_full_rounds - 1;
+    for i in 0..end {
+        let next_round = round_keys(i + 1);
+        let inverted = inverse_matrix.right_apply(next_round);
+        res.extend(inverted);
+    }
+
+    // The plan:
```
tsunrise, line 57: Still a plan? If not remove this

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_constants.rs
```diff
@@ -0,0 +1,39 @@
...
+use super::lfsr::GrainLFSR;
+use crate::crypto::hash::ParamField;
+use alloc::vec::Vec;
+
+/// return round constants, and return the LFSR used to generate MDS matrix
+pub fn generate_round_constants<F>(
+    prime_num_bits: u64,
+    width: usize,
+    r_f: usize,
+    r_p: usize,
```
tsunrise, line 28:
```suggestion
    num_full_rounds: usize,
    num_partial_rounds: usize
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_constants.rs
```diff
@@ -0,0 +1,39 @@
...
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Round Constants Generation
+
+use super::lfsr::GrainLFSR;
+use crate::crypto::hash::ParamField;
+use alloc::vec::Vec;
+
+/// return round constants, and return the LFSR used to generate MDS matrix
```
tsunrise, line 23:
```suggestion
/// Generates round constants, and the LFSR used to generate MDS matrix
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_numbers.rs
```diff
@@ -0,0 +1,140 @@
...
+//! Generate number of full & partial rounds
+//! Adapted from <https://github.com/filecoin-project/neptune/blob/master/src/round_numbers.rs>
+
+// The number of bits of the Poseidon prime field modulus. Denoted `n` in the Poseidon paper
+// (where `n = ceil(log2(p))`). Note that BLS12-381's scalar field modulus is 255 bits, however we
+// use 256 bits for simplicity when operating on bytes as the single bit difference does not affect
+// the round number security properties.
+const PRIME_BITLEN: usize = 256;
+
+// Security level (in bits), denoted `M` in the Poseidon paper.
```
tsunrise, line 26:
```suggestion
/// Security level (in bits), denoted `M` in the Poseidon paper.
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_numbers.rs
```diff
@@ -0,0 +1,140 @@
...
+    // Increase by 25%, rounding up.
+    let strengthened_partial_rounds = f64::ceil(partial_rounds as f64 * 1.25) as usize;
+
+    (full_round, strengthened_partial_rounds)
+}
+
+/// Returns the round numbers for a given width `t`. Here, the `security_margin` parameter does not
+/// indicate that we are calculating `R_F` and `R_P` for the "strengthened" round numbers, done in
+/// the function `round_numbers_strengthened()`.
+pub fn calc_round_numbers(t: usize, security_margin: bool) -> (usize, usize) {
```
tsunrise, line 61: Dead code? Round numbers are hardcoded in `config.rs`, and this file is no longer used.

tsunrise (changes requested):

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_numbers.rs
bhgomes: This still tells us what the safety margin is. We should add a test to check that the values defined in the spec match this (since we can't run this function as a `const-expr` yet).

bhgomes: This still tells us what the safety margin is. We should add a test to check that the values defined in the spec match this (since we can't run this function as a `const-expr` yet).

manta-pay/src/crypto/hash/poseidon/parameter_generation/lfsr.rs
```diff
@@ -0,0 +1,195 @@
...
+                .collect::<Vec<u8>>();
+
+            res.push(F::from_le_bytes_mod_order(&bytes));
+        }
+
+        res
+    }
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
BoyuanFeng: `arkworks` is included in `dev-dependencies` (https://github.com/Manta-Network/manta-rs/blob/main/manta-pay/Cargo.toml#L142).
So we do not need `#[cfg(feature = "arkworks")]` for `test`.

manta-pay/src/crypto/hash/poseidon/parameter_generation/preprocessing.rs
BoyuanFeng: I changed the name from "the plan" to "high-level workflow description". Since the computation here is complex, we need this description as an intuitive explanation of what is happening.

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_numbers.rs
BoyuanFeng: I added a test for that.

manta-pay/src/crypto/hash/poseidon/parameter_generation/mod.rs
BoyuanFeng: All of these mods will go to OpenZL and we would like to expose as much as possible.

manta-pay/src/crypto/hash/poseidon/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Poseidon Hash Implementation
+
+pub mod hasher;
+pub mod parameter_generation;
+
+/// Trait for Public Parameters in Hash Functions.
```
tsunrise, line 22: Can you clarify what are the "Public Parameters"? Are they round constants and MDS? If so mention them here.

tsunrise (commented):
nice work! left some minor comments

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,905 @@
...
+        Self::Scalar: Clone;
+
+    /// Returns an identity matrix of size `n*n`.
+    fn identity(n: usize) -> Self;
+
+    /// Multiplies matrix `self` with matrix `other` on the right side.
+    fn matmul(&self, other: &Self) -> Option<Self>
+    where
+        Self: Sized,
+        Self::Scalar: Clone;
```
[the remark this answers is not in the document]
bhgomes: The current algorithm generates an owned transpose matrix from a shared reference (so it requires a clone).

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,905 @@
...
+    }
+}
+
+impl<F> Index<usize> for Matrix<F>
+where
+    F: Field,
+{
+    type Output = Vec<F>;
+
+    /// Returns an unmutable reference to the `index`^{th} row in the matrix
```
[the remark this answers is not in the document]
bhgomes: IMO, the comment should be removed entirely, the documentation is already generated for implementations of traits. The comment can sometimes be useful, but only if the implementation does something surprising or the implementation of the trait is non-obvious from thinking of the trait API alone.

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+    where
+        F: Clone,
+    {
+        let mut shadow = Self::identity(self.num_rows());
+        self.upper_triangular(&mut shadow)?
+            .reduce_to_identity(&mut shadow)?;
+        Some(shadow)
+    }
+
+    /// Checks if the matrix is invertible
```
bhgomes, line 363:
```suggestion
    /// Checks if the matrix is invertible.
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SquareMatrix<F>(Matrix<F>)
+where
+    F: Field;
+
+impl<F> SquareMatrix<F>
+where
+    F: Field,
+{
+    /// Returns a new square matrix
```
bhgomes, line 347:
```suggestion
    /// Returns a new [`SquareMatrix`] representation of `m` if it returns `true` to [`is_square`](Matrix::is_square).
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+    fn to_row_major(self) -> Vec<F> {
+        self.0.to_row_major()
+    }
+
+    fn transpose(self) -> Self {
+        Self(self.0.transpose())
+    }
+}
+
+/// Inner product of two vectors.
```
bhgomes, line 518: Use present tense language, try to answer the question "What does this function do?"

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+pub fn inner_product<F>(a: &[F], b: &[F]) -> F
+where
+    F: Field,
+{
+    a.iter()
+        .zip(b)
+        .fold(F::zero(), |acc, (v1, v2)| F::add(&acc, &F::mul(v1, v2)))
+}
+
+/// Elementwise addition (i.e., out_i = a_i + b_i).
```
bhgomes, line 528: Replace "elementwise" with "pointwise" or just say vector addition.

Use present tense language, try to answer the question "What does this function do?"

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+pub fn vec_add<F>(a: &[F], b: &[F]) -> Vec<F>
+where
+    F: Field,
+{
+    a.iter().zip(b).map(|(a, b)| F::add(a, b)).collect()
+}
+
+/// Elementwise subtraction (i.e., out_i = a_i - b_i).
```
bhgomes, line 536: Replace "elementwise" with "pointwise" or just say vector subtraction.

Use present tense language, try to answer the question "What does this function do?"

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+
+/// Eliminates `row` with `factor` multiplying `pivot`.
+fn eliminate_row<F>(row: &[F], factor: &F, pivot: &[F]) -> Vec<F>
+where
+    F: Field,
+{
+    vec_sub(row, &scalar_vec_mul(factor, pivot))
+}
+
+/// Returns kronecker delta.
```
bhgomes, line 560:
```suggestion
/// Returns the kronecker delta of `i` and `j`.
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+    }
+}
+
+/// Checks whether `elem` equals zero.
+pub fn equal_zero<F>(elem: &F) -> bool
+where
+    F: Field,
+{
+    F::eq(elem, &F::zero())
+}
```
bhgomes, line 578: Is this function used? If it is, then add it to the `Field` trait and call it `is_zero`.

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+        ];
+        let m = SquareMatrix::new(m).unwrap();
+        for (i, j, expected) in &cases {
+            let result = m.minor(*i, *j).unwrap();
+            assert_eq!(expected, &result);
+        }
+    }
+
+    #[test]
+    fn scalar_mul_is_correct() {
```
bhgomes, line 660: Add a doc string.

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
```diff
@@ -0,0 +1,900 @@
...
+        let m = Matrix::new(vec![vec![zero, one], vec![two, three]]).unwrap();
+        let res = m.mul_by_scalar(two);
+
+        let expected = Matrix::new(vec![vec![zero, two], vec![four, six]]).unwrap();
+
+        assert_eq!(expected.0, res.0);
+    }
+
+    #[test]
+    fn vec_mul_is_correct() {
```
bhgomes, line 677: Add a doc string.

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_numbers.rs
```diff
@@ -0,0 +1,170 @@
...
+    ]
+    .iter()
+    .map(|num_full_rounds| num_full_rounds.ceil() as usize)
+    .max()
+    .unwrap();
+    num_full_rounds >= num_full_rounds_max
+}
+
+#[cfg(test)]
+mod test {
```
bhgomes, line 121: Add doc string.

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_numbers.rs
```diff
@@ -0,0 +1,170 @@
...
+    use super::*;
+    use crate::{config::PoseidonSpec, crypto::hash::poseidon::hasher::arkworks::Specification};
+
+    #[test]
+    fn round_numbers_matches_known_values() {
```
bhgomes, line 126: Add doc string.

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_numbers.rs
```diff
@@ -0,0 +1,170 @@
...
+            (65, 61),
+        ];
+        for (t, rp_expected) in cases.iter() {
+            let (rf, rp) = calc_round_numbers(*t, true);
+            assert_eq!(rf, 8);
+            assert_eq!(rp, *rp_expected);
+        }
+    }
+
+    fn compare_against_config_values<const ARITY: usize>()
```
bhgomes, line 156: Add doc string.

manta-pay/src/crypto/hash/poseidon/parameter_generation/round_numbers.rs
```diff
@@ -0,0 +1,170 @@
...
+    where
+        PoseidonSpec<ARITY>: Specification,
+    {
+        let (num_full_rounds, num_partial_rounds) = calc_round_numbers(ARITY + 1, true);
+        assert_eq!(num_full_rounds, PoseidonSpec::<ARITY>::FULL_ROUNDS);
+        assert_eq!(num_partial_rounds, PoseidonSpec::<ARITY>::PARTIAL_ROUNDS);
+    }
+
+    #[test]
```
bhgomes, line 165: Add doc string.

manta-util/src/vec.rs
```diff
@@ -38,6 +40,17 @@ pub trait VecExt<T>: Into<Vec<T>> + sealed::Sealed + Sized {
...
+
+    /// Allocates a vector of length `n` and initializes with `f`.
+    #[inline]
+    fn allocate_with<F>(n: usize, f: F) -> Self
+    where
+        F: FnMut() -> T,
+    {
+        let mut v = Vec::with_capacity(n);
+        v.resize_with(n, f);
+        v.into()
```
bhgomes, line 52:
```suggestion
        let mut vec = Vec::with_capacity(n);
        vec.resize_with(n, f);
        vec.into()
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! MDS Data Generation
+
+use super::matrix::{MatrixOperations, SquareMatrix};
+use crate::crypto::hash::poseidon::{parameter_generation::matrix::Matrix, Field, FieldGeneration};
```
bhgomes, line 20: merge these

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+    pub m_hat_inv: SquareMatrix<F>,
+    /// m prime matrix. Used in optimized poseidon hash.
+    pub m_prime: SquareMatrix<F>,
+    /// m double prime matrix. Used in optimized poseidon hash.
+    pub m_double_prime: SquareMatrix<F>,
+}
+
+impl<F> MdsMatrices<F>
+where
+    F: Field + Clone,
```
bhgomes, line 49: order

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+            .skip(1)
+            .map(|column| column[0].clone())
+            .collect();
+        (v, w)
+    }
+}
+
+impl<F> MdsMatrices<F>
+where
+    F: Field + Clone,
```
bhgomes, line 64: order

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+                            let mut new_row = Vec::allocate_with(row.len(), F::zero);
+                            new_row[1..].clone_from_slice(&row[1..]);
+                            new_row
+                        }
+                    })
+                    .collect(),
+            )
+            .unwrap(),
+        )
+        .unwrap()
```
bhgomes, line 87: replace with expect

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+}
+
+impl<F> MdsMatrices<F>
+where
+    F: Clone + Field + FieldGeneration + PartialEq,
```
bhgomes, line 93: minimize requirements

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+    pub fn new(dim: usize) -> Option<Self> {
+        Self::generate_mds(dim).map(Self::derive_mds_matrices)
+    }
+
+    /// Generates the mds matrix `m` for naive poseidon hash.
+    /// mds matrix is constructed to be symmetry so that row-major or col-major
+    /// representation gives the same output
+    pub fn generate_mds(t: usize) -> Option<SquareMatrix<F>>
+    where
+        F: FieldGeneration,
```
bhgomes, line 105: minimize requirements

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+                            new_row[0] = w_hat[i - 1].clone();
+                            new_row[i] = F::one();
+                            new_row
+                        }
+                    })
+                    .collect(),
+            )
+            .unwrap(),
+        )
+        .unwrap()
```
bhgomes, line 153: use expect

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+                    .map(|x| {
+                        ys.iter()
+                            .map(|y| F::add(&F::from_u64(x), y).inverse().unwrap())
+                            .collect()
+                    })
+                    .collect(),
+            )
+            .unwrap(),
+        )
+        .unwrap();
```
bhgomes, line 120: use expect

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+            m_prime,
+            m_double_prime,
+        }
+    }
+}
+
+/// A `SparseMatrix` is specifically one of the form of M''.
+/// This means its first row and column are each dense, and the interior matrix
+/// (minor to the element in both the row and column) is the identity.
+/// We will pluralize this compact structure `sparse_matrixes` to distinguish from `sparse_matrices` from which they are created.
```
bhgomes, line 177: what does "pluralize" mean

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+    /// `v_rest` contains all but the first (already included in `w_hat`).
+    pub v_rest: Vec<F>,
+}
+
+impl<F> SparseMatrix<F>
+where
+    F: Field,
+{
+    /// Checks if `self` is square and `self[1..][1..]` is identity
+    fn is_sparse(m: &SquareMatrix<F>) -> bool
```
bhgomes, line 194:
```suggestion
    pub fn is_sparse(m: &SquareMatrix<F>) -> bool
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+            None => false,
+        }
+    }
+
+    /// Generates sparse matrix from m_double_prime matrix
+    pub fn new(m_double_prime: SquareMatrix<F>) -> Self
+    where
+        F: Clone,
+    {
+        assert!(Self::is_sparse(&m_double_prime));
```
bhgomes, line 209: return `Option` instead

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+    {
+        let mut m = Matrix::identity(self.size());
+        for (j, elt) in self.w_hat.iter().enumerate() {
+            m[j][0] = elt.clone();
+        }
+        for (i, elt) in self.v_rest.iter().enumerate() {
+            m[0][i + 1] = elt.clone();
+        }
+        m
+    }
```
bhgomes, line 234:
```suggestion
    /// Generates dense-matrix representation from sparse matrix representation
    pub fn to_matrix(self) -> Matrix<F> {
        let mut matrix = Matrix::identity(self.size());
        for (j, elem) in self.w_hat.into_iter().enumerate() {
            matrix[j][0] = elem;
        }
        for (i, elem) in self.v_rest.into_iter().enumerate() {
            matrix[0][i + 1] = elem;
        }
        matrix
    }
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// Factorizes into sparse matrices.
```
bhgomes, line 237: factorizes what? what does this mean

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+pub fn factor_to_sparse_matrixes<F>(
+    base_matrix: SquareMatrix<F>,
+    n: usize,
+) -> (SquareMatrix<F>, Vec<SparseMatrix<F>>)
+where
+    F: Clone + Field + FieldGeneration + PartialEq,
+{
+    let (pre_sparse, mut sparse_matrices) =
+        (0..n).fold((base_matrix.clone(), Vec::new()), |(curr, mut acc), _| {
```
bhgomes, line 246: Do we know the size of this `Vec`?

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+            let new = base_matrix.matmul(&derived.m_prime).unwrap();
+            (new, acc)
+        });
+    sparse_matrices.reverse();
+    let sparse_matrices = sparse_matrices.into_iter().map(SparseMatrix::new).collect();
+    (pre_sparse, sparse_matrices)
+}
+
+#[cfg(test)]
+mod test {
```
bhgomes, line 258: add docs

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+    use super::*;
+    use ark_bls12_381::Fr;
+    use ark_ff::field_new;
+    use ark_std::{test_rng, UniformRand};
+
+    use crate::crypto::{
+        constraint::arkworks::Fp, hash::poseidon::parameter_generation::matrix::Matrix,
+    };
```
bhgomes, line 266:
```suggestion
    use super::*;
    use ark_bls12_381::Fr;
    use ark_ff::field_new;
    use ark_std::{test_rng, UniformRand};
    use crate::crypto::{
        constraint::arkworks::Fp, hash::poseidon::parameter_generation::matrix::Matrix,
    };
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+
+    #[test]
+    fn mds_matrices_creation_is_correct() {
```
bhgomes, line 269: add doc string

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+        for i in 2..5 {
+            mds_matrices_creation_aux(i);
+        }
+    }
+
+    fn mds_matrices_creation_aux(width: usize) {
```
bhgomes, line 275: give a better name for this. What does this do?

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        // M^-1 x M = I
+        assert!(m_inv.matmul(&m).unwrap().is_identity());
+
+        // M' x M'' = M
+        assert_eq!(m, m_prime.matmul(&m_double_prime).unwrap());
+    }
+
+    #[test]
+    fn swapping_is_correct() {
```
bhgomes, line 299: add doc string

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+        swapping_aux(3)
+    }
+
+    fn swapping_aux(width: usize) {
```
bhgomes, line 303: add doc string, change name, remove whitespace, and remove type annotations where possible

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+        let x_m1_m2 = mds
+            .m_double_prime
+            .mul_row_vec_at_left(&mds.m_prime.mul_row_vec_at_left(&x).unwrap())
+            .unwrap();
+        assert_eq!(xm, x_m1_m2);
+    }
+
+    #[test]
+    fn mds_matches_hardcoded_sage_output() {
+        // value come out from sage script
```
bhgomes, line 344:
```suggestion
        // value come out from sage script
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+        let xm = mds.m.mul_row_vec_at_left(&x).unwrap();
+        let x_m1_m2 = mds
+            .m_double_prime
+            .mul_row_vec_at_left(&mds.m_prime.mul_row_vec_at_left(&x).unwrap())
+            .unwrap();
+        assert_eq!(xm, x_m1_m2);
+    }
+
+    #[test]
+    fn mds_matches_hardcoded_sage_output() {
```
bhgomes, line 343: add doc string

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+                    Fr,
+                    "43696562645938492066206450423488304864742127083773031518836382249948817653761"
+                )),
+                Fp(field_new!(
+                    Fr,
+                    "14981678621464625851270783002338847382197300714436467949315331057125308909861"
+                )),
+            ],
+        ])
+        .unwrap();
```
bhgomes, line 391: use expect

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        let mds = MdsMatrices::<Fp<Fr>>::generate_mds(width).unwrap();
+        assert_eq!(mds, expected_mds);
```
bhgomes, line 394:
```suggestion
        assert_eq!(MdsMatrices::generate_mds(width).expect("For width=3, we can always generate a valid MDS matrix."), expected_mds);
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    #[test]
+    fn mds_is_invertible() {
```
bhgomes, line 398: add doc string

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+        for t in 3..10 {
+            let mds = MdsMatrices::<Fp<Fr>>::generate_mds(t).unwrap();
+            assert!(mds.is_invertible());
```
bhgomes, line 401:
```suggestion
            assert!(MdsMatrices::<Fp<Fr>>::generate_mds(t).expect("Failed to generate MDS matrix.").is_invertible());
```

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
```diff
@@ -0,0 +1,412 @@
...
+        }
+    }
+
+    #[test]
+    fn mds_is_symmetric() {
```
bhgomes, line 406: add docs

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -19,46 +19,62 @@
...
+    /// Field used as state
     type Field;
 
-    /// Number of Full Rounds
-    ///
-    /// This is counted twice for the first set of full rounds and then the second set after the
-    /// partial rounds.
-    const FULL_ROUNDS: usize;
+    /// Field used as constants
+    type ParameterField;
```
bhgomes, line 45:
```suggestion
    /// Field Type used for Constant Parameters
    type ParameterField;
```

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -19,46 +19,62 @@
...
 
     /// Number of Partial Rounds
     const PARTIAL_ROUNDS: usize;
 
-    /// Returns the additive identity of the field.
-    fn zero(compiler: &mut COM) -> Self::Field;
+    /// Number of Full Rounds
+    ///
+    /// The total number of full rounds in Poseidon hash, including the first set
```
bhgomes, line 52:
```suggestion
    /// The total number of full rounds in Poseidon Hash, including the first set
```

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -19,46 +19,62 @@
...
+    /// of full rounds and then the second set after the partial rounds.
+    const FULL_ROUNDS: usize;
+
+    /// Returns the domain tag. We use different domain_tags different applications
+    /// so that defending against rainbow table attack.
+    fn domain_tag(arity: usize, compiler: &mut COM) -> Self::Field;
```
bhgomes, line 58:
```suggestion
    /// Returns the domain tag for `arity`. We use different domain tags for different applications
    /// to defend against rainbow table attacks.
    fn domain_tag(arity: usize, compiler: &mut COM) -> Self::Field;
```

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -114,8 +130,18 @@ where
     /// Width of the State Buffer
     pub const WIDTH: usize = ARITY + 1;
 
+    /// Half Number of Full Rounds
+    ///
+    /// Poseidon hash first has [`HALF_FULL_ROUNDS`] full rounds in the beginning,
+    /// followed by [`PARTIAL_ROUNDS`] partial rounds in the middle, and finally
+    /// [`HALF_FULL_ROUNDS`] full rounds at the end.
```
bhgomes, line 137:
```suggestion
    /// Poseidon Hash first has [`HALF_FULL_ROUNDS`]-many full rounds in the beginning,
    /// followed by [`PARTIAL_ROUNDS`]-many partial rounds in the middle, and finally
    /// [`HALF_FULL_ROUNDS`]-many full rounds at the end.
```

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -245,41 +281,42 @@ where
...
-    ///
-    /// This method samples the individual field elements of the parameters set, instead of
-    /// producing an actually correct/safe set of additive round keys and MDS matrix.
     #[inline]
     fn sample<R>(distribution: D, rng: &mut R) -> Self
     where
         R: CryptoRng + RngCore + ?Sized,
     {
+        let _ = distribution;
+        let _ = rng;
```
bhgomes, line 297:
```suggestion
        let _ = (distribution, rng);
```

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -329,10 +366,14 @@ pub type Output<S, COM, const ARITY: usize> =
...
-    use ark_ff::{Field, PrimeField};
-    use ark_r1cs_std::fields::FieldVar;
-    use manta_crypto::constraint::{Constant, ValueSource};
+    use crate::crypto::{
+        constraint::arkworks::{Fp, FpVar, R1CS},
+        hash::poseidon::FieldGeneration,
+    };
+    use ark_ff::{BigInteger, Field, FpParameters, PrimeField};
+    use ark_r1cs_std::{alloc::AllocVar, fields::FieldVar};
+    use ark_std::{One, Zero};
```
bhgomes, line 375: Don't use `ark_std`, use `ark_ff` for these imports.

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -344,8 +385,8 @@ pub mod arkworks {
 
         /// Number of Full Rounds
         ///
-        /// This is counted twice for the first set of full rounds and then the second set after the
-        /// partial rounds.
+        /// The total number of full rounds in Poseidon hash, including the first set
```
bhgomes, line 388:
```suggestion
        /// The total number of full rounds in Poseidon Hash, including the first set
```

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -355,35 +396,103 @@ pub mod arkworks {
...
+    where
+        F: PrimeField,
+    {
+        fn zero() -> Self {
+            Self(<F as Zero>::zero())
+        }
+
+        fn one() -> Self {
+            Self(<F as One>::one())
+        }
```
bhgomes, line 409: Can you use `F::zero` and `F::one` here?

manta-pay/src/crypto/hash/poseidon/hasher.rs
```diff
@@ -395,37 +504,61 @@ pub mod arkworks {
...
         const PARTIAL_ROUNDS: usize = S::PARTIAL_ROUNDS;
+        const FULL_ROUNDS: usize = S::FULL_ROUNDS;
 
         #[inline]
-        fn zero(compiler: &mut Compiler<S>) -> Self::Field {
-            let _ = compiler;
-            Self::Field::zero()
+        fn domain_tag(arity: usize, compiler: &mut Compiler<S>) -> Self::Field {
+            let v = S::Field::from(((1 << arity) - 1) as u64);
+            FpVar::new_witness(compiler.cs.clone(), || Ok(v)).unwrap()
```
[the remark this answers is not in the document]
bhgomes: Should we make this a constant?

manta-pay/src/crypto/hash/poseidon/parameter_generation/matrix.rs
BoyuanFeng: we should use elementwise instead of pointwise, since `elementwise` is more commonly used: https://www.google.com/search?q=elementwise+addition&oq=elementwise+addition&aqs=chrome..69i57j0i10i512l3j0i512j0i10i512l2j0i22i30l3.2674j0j7&sourceid=chrome&ie=UTF-8

manta-pay/src/crypto/hash/poseidon/parameter_generation/mds.rs
BoyuanFeng: `FieldGeneration` is required for `F::from_u64`.

manta-pay/src/crypto/hash/poseidon/mod.rs
```diff
@@ -172,7 +244,7 @@ where
             let linear_combination = state
```
tsunrise, line 244: Can you update the comment above that explains clippy is wrong?

```
// NOTE: Clippy is wrong here. Without `collect`, the two closures in `map` and `reduce` will have simultaneous `&mut` access to compiler. Adding `collect` allows `map` to be done first before `reduce`.
```

manta-pay/src/crypto/hash/poseidon/parameter/round_numbers.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// Testing Suite
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::{config::PoseidonSpec, crypto::hash::poseidon::arkworks::Specification};
+
+    /// Checks if values from `calc_round_numbers` matches known values.
+    #[test]
+    fn round_numbers_matches_known_values() {
```
tsunrise, line 127: nit `round_numbers_match_known_values`

tsunrise, line 127: nit: `round_numbers_match_known_values`

manta-pay/src/crypto/hash/poseidon/parameter/round_numbers.rs
```diff
@@ -0,0 +1,172 @@
...
+        PoseidonSpec<ARITY>: Specification,
+    {
+        let (num_full_rounds, num_partial_rounds) = calc_round_numbers(ARITY + 1, true);
+        assert_eq!(num_full_rounds, PoseidonSpec::<ARITY>::FULL_ROUNDS);
+        assert_eq!(num_partial_rounds, PoseidonSpec::<ARITY>::PARTIAL_ROUNDS);
+    }
+
+    /// Checks if config values matches values from `calc_round_numbers`
+    #[test]
+    fn round_numbers_matches_config_values() {
```
tsunrise, line 168: nit: `round_numbers_match_config_values`

tsunrise (approved):
LGTM with some minor comments. I pre-approve this so that my review will not block this PR, but do not merge until @bhgomes takes another round

manta-pay/src/crypto/hash/poseidon/parameter/matrix.rs
```diff
@@ -0,0 +1,887 @@
...
+        Self::Scalar: Clone;
+
+    /// Elementwisely multiplies with `scalar`.
+    fn mul_by_scalar(&self, scalar: Self::Scalar) -> Self;
+
+    /// Returns row major representation of the matrix.
+    fn to_row_major(self) -> Vec<Self::Scalar>;
+
+    /// Returns the transpose of the matrix.
+    fn transpose(self) -> Self;
```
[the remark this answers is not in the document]
BoyuanFeng: If we use `&self` here, we need an additional trait bound `Clone`. Since `self` will be cloned anyway, we pass by value instead of pass by reference.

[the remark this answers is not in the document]
tsunrise: We only need `Clone` trait for the field, no? If we want still want to use the original matrix, this will cost one extra clone?

bhgomes (approved):

tsunrise (approved):
LGTM

manta-pay/src/crypto/hash/poseidon/mod.rs
```diff
@@ -0,0 +1,636 @@
...
+        #[inline]
+        fn new_constant(this: &Self::Type, compiler: &mut Compiler<S>) -> Self {
+            let _ = compiler;
+            Self {
+                additive_round_keys: this.additive_round_keys.clone(),
+                mds_matrix: this.mds_matrix.clone(),
+            }
+        }
+    }
+}
```
tsunrise, line 636: I don't see any hardcoded test for Poseidon here? How to ensure this implementation is correct?

bhgomes (approved):

PR #81, opened 2022-05-26
DO NOT MERGE: feat: upgrade to new crypto primitives

bhgomes:
**CLOSE WHEN #149 IS MERGED**

---

We need to add the following cryptographic primitives to get ready for the v1.0.0 release:

- [x] Authenticated Encryption @bhgomes 
- [x] Duplex Sponge Method @tsunrise @BoyuanFeng @bhgomes 
- [x] Poseidon as Permutation @tsunrise @BoyuanFeng @bhgomes 
- [x] Poseidon as Sponge @tsunrise @BoyuanFeng @bhgomes 
- [x] Signature Scheme @bhgomes @BoyuanFeng
    - [x] Signature Abstraction
    - [x] Schnorr Implementation
- [x] ECLAIR refactor (dog-fooding) @bhgomes

---

- [ ] Added **one** line describing your change in [`CHANGELOG.md`](https://github.com/manta-network/manta-rs/blob/main/CHANGELOG.md) and added the appropriate `changelog` label to the PR.
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer.


manta-crypto/src/mac.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// Message Authentication Code
+pub trait MessageAuthenticationCode<COM = ()> {
+    /// Key Type
+    type Key: ?Sized;
+
+    /// Message Type
+    type Message: ?Sized;
+
+    /// Digest Type
```
BoyuanFeng, line 30: The name here seems to be different from the spec. E.g., `Digest` v.s. `Tag`. Should we make it consistent?

manta-crypto/src/mac.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    /// Computes the message authentication code against `key` and `message`.
+    #[inline]
+    fn hash(&self, key: &Self::Key, message: &Self::Message) -> Self::Digest
+    where
+        COM: Native,
+    {
+        self.hash_with(key, message, &mut COM::compiler())
+    }
+}
```
BoyuanFeng, line 49: Why don't we have a `verify` function in this trait?

manta-crypto/src/mac.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn hash_with(
+        &self,
+        key: &Self::Key,
+        message: &Self::Message,
+        compiler: &mut COM,
+    ) -> Self::Digest;
+
+    /// Computes the message authentication code against `key` and `message`.
+    #[inline]
+    fn hash(&self, key: &Self::Key, message: &Self::Message) -> Self::Digest
```
BoyuanFeng, line 43: Should we use `sign` instead of `hash` to match spec?

BoyuanFeng (changes requested):

manta-crypto/Cargo.toml
```diff
@@ -25,22 +25,25 @@ is-it-maintained-open-issues = { repository = "Manta-Network/manta-rs" }
...
 # Serde
-serde = ["manta-util/serde-alloc", "manta-util/serde-array"]
+serde = ["manta-util/serde-array"]
 
 # Standard Library
 std = ["manta-util/std"]
 
 # Testing Frameworks
-test = []
+test = ["alloc"]
```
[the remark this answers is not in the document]
bhgomes: We need `alloc` for `merkle_tree`.

manta-pay/src/crypto/poseidon/encryption.rs
```diff
@@ -0,0 +1,191 @@
...
+    S::Field: Clone,
+{
+    type Output = CiphertextBlock<S, COM>;
+
+    #[inline]
+    fn write(&self, state: &mut State<S, COM>, compiler: &mut COM) -> Self::Output {
+        for (i, elem) in state.iter_mut().skip(1).enumerate() {
+            *elem = S::add(elem, &self.0[i], compiler);
+        }
+        CiphertextBlock(state.iter().skip(1).cloned().collect())
```
tsunrise, line 68: I'm not sure if this is the correct ciphertext block. In the paper, we XOR/ADD the plaintext to the state (correspond to the for-loop here), do a permute, and the ciphertext is plaintext XOR/ADD the state after permute. However, the permutation is missing here? Correct me if wrong

bhgomes: (Reading line numbers from the paper).

On line 15 we start by first computing the ciphertext block against the current state (`Z` is the current state without capacity) then we call `duplexing` against `B_0` which again XORs with the state (without capacity). (We are not using frame bits in our algorithm, but the `duplex` API supports frame bits if people want them). So instead we collapse these two operations into the API called `write` which produces the ciphertext block `C_0` and updates the state (the first half of `duplexing`). Then, we immediately permute the state in the `duplex_encryption` function which is the second-half of the `duplexing` call on line 17. Then we loop at that point instead of looping on line 18. We shift the loop by one instruction which still gives the same result (line 15 becomes part of the first loop instead of some instruction before the loop starts and line 20 is also just part of the loop, the last instruction in the last iteration).

manta-crypto/src/eclair/alloc.rs
```diff
@@ -0,0 +1,305 @@
...
+pub trait Variable<M, COM>
+where
+    COM: ?Sized,
+{
+    /// Underlying Type
+    type Type;
+
+    /// Allocates a new unknown value into the `compiler`. The terminology "unknown" refers to the
+    /// fact that we need to allocate a slot for this variable during compilation time, but do not
+    /// yet know its underlying value.
```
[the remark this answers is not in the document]
bhgomes: Those two docs are only true for the execution engines that build contexts or proofs, but these execution engines can be more general, so those are not necessarily correct. The only distinction that allocation makes is between the values being known at compilation time or execution time. Some execution engines don't need to know the values in which case they just use the unknown allocation slots (this corresponds with building proving/verifying contexts in that case).

manta-crypto/src/permutation/duplex.rs
```diff
@@ -0,0 +1,310 @@
...
+    fn initialize(&self, compiler: &mut COM) -> P::Domain;
+
+    /// Generates the starting input blocks for `key` and `header` data to be inserted into the
+    /// cipher.
+    fn setup(
+        &self,
+        key: &Self::Key,
+        header: &Self::Header,
+        compiler: &mut COM,
+    ) -> Vec<Self::SetupBlock>;
```
BoyuanFeng, line 83: Should we change `Vec` here to `box`?

manta-crypto/src/permutation/sponge.rs
```diff
@@ -0,0 +1,108 @@
...
+        Self { permutation, state }
+    }
+
+    /// Updates `self` by absorbing writes into the state with `input`.
+    #[inline]
+    pub fn absorb<W>(&mut self, input: &W, compiler: &mut COM) -> W::Output
+    where
+        W: Write<P, COM>,
+    {
+        let out = input.write(self.state, compiler);
```
BoyuanFeng, line 78: Should we make`input.write()` code style consistent with the `R::read(self.state, compiler)` below?

BoyuanFeng (changes requested):

manta-pay/src/crypto/poseidon/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+    S: Specification<COM>;
+
+impl<S, COM> State<S, COM>
+where
+    S: Specification<COM>,
+{
+    /// Builds a new [`State`] from `state`.
+    #[inline]
+    pub fn new(state: Box<[S::Field]>) -> Self {
+        assert_eq!(state.len(), S::WIDTH);
```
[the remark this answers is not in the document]
bhgomes: How does it make it slower? Using `None` here is slower.

manta-pay/src/crypto/poseidon/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+                .iter()
+                .enumerate()
+                .map(|(j, elem)| S::mul_const(elem, &self.mds_matrix[S::WIDTH * i + j], compiler))
+                .collect::<Vec<_>>();
+            next.push(
+                linear_combination
+                    .into_iter()
+                    .reduce(|acc, next| S::add(&acc, &next, compiler))
+                    .unwrap(),
+            );
```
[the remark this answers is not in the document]
bhgomes: This doesn't work because of the double borrow of `compiler`. See the `clippy` warning above.

manta-pay/src/crypto/poseidon/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+    }
+}
+
+impl<D, S, COM> Sample<D> for Permutation<S, COM>
+where
+    D: Clone,
+    S: Specification<COM>,
+    S::ParameterField: Field + FieldGeneration + PartialEq + Sample<D>,
+{
+    /// Samples random Poseidon parameters.
```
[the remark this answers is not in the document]
bhgomes: Yes.

PR #82, opened 2022-05-28
Implement randomizable key derivation function

BoyuanFeng:


manta-crypto/src/ecc.rs
```diff
@@ -334,6 +334,36 @@ where
...
+    fn rand_output_in(
+        &self,
+        randomness: &Self::Randomness,
+        output: &Self::Output,
+        compiler: &mut COM,
+    ) -> Self::Output {
+        output.scalar_mul(randomness, compiler)
+    }
+}
+
```
BoyuanFeng, line 366: TODO:
1. Add a test on the `Random Derivation` property.
2. The [test for key derivation function](https://github.com/Manta-Network/manta-rs/pull/82/files#diff-0a2f47f9702f7f522d33e2efcc120d9b919caf59c7ca309f8f04719f57cc3c7bR619-R629) is never called. We may need to also add a test for it.

bhgomes:
Closed due to removal from spec.

PR #83, opened 2022-05-28
Feature: Schnorr Signature

BoyuanFeng:


bhgomes:
We also need to add a signature test for the signature property.

bhgomes:
Closed in favor of #81.

PR #129, opened 2022-06-22
fix: reduce cost of signer key-search algorithm with pre-compute table

bhgomes:

Before this PR, the signer key-search algorithm would re-compute every viewing key within the search range on every insertion of a UTXO. Now, we use a dynamic program that replaces the viewing key computation with a lookup if the key has already been computed. To prevent run-away growth of this table we have a max capacity set to `16` times the `GAP_LIMIT` (currently set to `20`), a very unlikely limit to be reached except for implementation errors or purposeful memory attacks.

This PR also adds some utility functions for B-Trees.

Focus Files:

- `manta-accounting/src/wallet/signer.rs`
- `manta-accounting/src/key.rs`

---


tjsharp1:
out

manta-accounting/src/key.rs
```diff
@@ -524,6 +523,71 @@ where
...
+    /// [`VIEW_KEY_BUFFER_LIMIT`]: Self::VIEW_KEY_BUFFER_LIMIT
+    #[inline]
+    pub fn view_key(&mut self, index: KeyIndex) -> Option<&H::SecretKey> {
+        btree_map::get_or_mutate(&mut self.view_keys, &index, |map| {
+            let next_key = self.keys.view_key(index)?;
+            if map.len() == Self::VIEW_KEY_BUFFER_LIMIT {
+                btree_map::pop_last(map);
+            }
+            map.insert(index, next_key);
+            map.get(&index)
```
BoyuanFeng, line 587: `map.get()` returns `next_key` and takes some time to search the btree. Could we somehow avoid this search cost?

manta-accounting/src/key.rs
```diff
@@ -612,6 +676,25 @@ where
     pub item: T,
 }
 
+impl<H, T> ViewKeySelection<H, T>
```
BoyuanFeng, line 679: 1. Can we rename `ViewKeySelection` as `IndexSelectedViewKey`? `ViewKeySelection` sounds like an action to select something with `ViewKey`. Instead, we are actually using `index` to find a `keypair` and `item` here.
2. Can we rename `keypair` to `key_pair`?

manta-accounting/src/wallet/signer.rs
```diff
@@ -630,63 +633,69 @@ where
...
-                        deposit.push(asset);
-                    }
-                    return Ok(());
+            },
+        } = selection;
+        if let Some(void_number) =
+            parameters.check_full_asset(&keypair.spend, &ephemeral_secret_key, &asset, &utxo)
+        {
+            if let Some(index) = void_numbers.iter().position(move |v| v == &void_number) {
+                void_numbers.remove(index);
```
BoyuanFeng, line 676: Why do we have `void_numbers.remove(index)`? `void_number` should be an append-only vector?

BoyuanFeng (changes requested):

manta-accounting/src/key.rs
bhgomes: Yeah, it might be possible to do `map.entry(index).or_insert(next_key)`. Let me try that.

manta-accounting/src/wallet/signer.rs
bhgomes: `void_numbers` is the set of void numbers during this synchronization round. Once we find the pre-image for a void number in this round, we can remove it from the set since it won't be matched again.

manta-accounting/src/key.rs
bhgomes: 1. The `index` is just a name for the view key (we don't refer to it by its actual secret value). We are still using the view key to select the item (i.e. can it be decrypted under the key), but the keypair is also returned for convenience (its not what's being searched for).
2. Yeah we can rename it, but it will be renamed everywhere then.

BoyuanFeng (approved):

Issue #114, opened 2022-06-21
Add Public Input Generation to `Transfer`


bhgomes:
Right now, only the `TransferPost` structures can return their data as a public input `struct`. We also want `Transfer` to be able to produce this vector too.

Old code in `struct TransferPost<C>` (manta-accounting/src/transfer/mod.rs):
```rust
    /// Generates the public input for the [`Transfer`] validation proof.
    #[inline]
    pub fn generate_proof_input(&self) -> ProofInput<C> {
        let mut input = Default::default();
        if let Some(asset_id) = self.asset_id {
            C::ProofSystem::extend(&mut input, &asset_id);
        }
        self.sources
            .iter()
            .for_each(|source| C::ProofSystem::extend(&mut input, source));
        self.sender_posts
            .iter()
            .for_each(|post| post.extend_input(&mut input));
        self.receiver_posts
            .iter()
            .for_each(|post| post.extend_input(&mut input));
        self.sinks
            .iter()
            .for_each(|sink| C::ProofSystem::extend(&mut input, sink));
        input
    }
```

We want something similar for Transfer where `self.sender_posts` and `self.receiver_posts` are changed to `self.sender` and `self.receiver`. Note that `extend_input` function needs to be implemented for `receiver` and `sender`.

PR #133, opened 2022-06-24
feat: Add generate-proof-input for Transfer

BoyuanFeng:
Adds public input generation to `Transfer`.

Closes: #114 

---


manta-accounting/src/transfer/mod.rs
```diff
@@ -1723,6 +1723,28 @@ where
         Self::new_unchecked(asset_id, sources, senders, receivers, sinks)
     }
 
+    /// Generates the public input for the [`Transfer`] validation proof.
+    #[inline]
+    pub fn generate_proof_input(self) -> ProofInput<C> {
```
bhgomes, line 1728: This should be the same API as `TransferPost` which borrows its input, not moves it.

bhgomes (changes requested):
Also, please add the compatibility test.

manta-accounting/src/transfer/mod.rs
BoyuanFeng: `into_post(self)` takes `self` as argument. We need to move `self` into `generate_proof_input`. Otherwise, we may need the `Clone` trait for `sender` and `receiver`.

bhgomes: The `into_post` method is not the one to use. Just borrow the data that `into_post` would move.

bhgomes: Remember in our conversation that you needed to add `extend_input` methods to `Sender` and `Receiver`.

manta-accounting/src/transfer/mod.rs
```diff
@@ -1396,6 +1403,12 @@ where
             encrypted_note: self.encrypted_note,
         }
     }
+
+    /// Extends proof public input with `self`
```
bhgomes, line 1407:
```suggestion
    /// Extends proof public input with `self`.
```

manta-accounting/src/transfer/test.rs
```diff
@@ -243,6 +245,36 @@ where
...
+                utxo_accumulator,
+            },
+            rng,
+        );
+        let full_parameters = FullParameters::new(parameters, utxo_accumulator.model());
+        let (proving_context, _) = Self::generate_context(public_parameters, full_parameters, rng)?;
+        Ok(transfer.generate_proof_input()
+            == transfer
+                .into_post(full_parameters, &proving_context, rng)
+                .expect("Unable to convert `Transfer` into `TransferPost`.")
```
bhgomes, line 275: You should be able to just use `?` here.

manta-pay/src/test/transfer.rs
```diff
@@ -103,6 +103,42 @@ fn reclaim() {
...
+    assert!(
+        Mint::sample_and_check_generate_proof_input_compatibility(
+            &(),
+            &rng.gen(),
+            &mut UtxoAccumulator::new(rng.gen()),
+            &mut rng
+        )
+        .expect("For a random Mint, `generate_proof_input` from `Transfer` and `ProofInput` should give the same `ProofInput`."),
+        "For a random Mint, `generate_proof_input` from `Transfer` and `TransferPost` should have given the same `ProofInput`."
+    );
```
bhgomes, line 119: You can write this as `assert!(matches!(..., Ok(true)), "...")`

bhgomes (changes requested):

bhgomes (approved):

tsunrise (approved):

PR #145, opened 2022-06-28
feat: add improvements to speed up CI builds

bhgomes:
- [x] Adds `cargo-hakari` for workspace hacks
- [x] Replaces `cargo test` with `cargo nextest`
- [x] Makes long-running jobs dependent on `format`

---


manta-workspace-hack/Cargo.toml
```diff
@@ -0,0 +1,66 @@
+# This file is generated by `cargo hakari`.
```
tsunrise, line 1: From https://crates.io/crates/cargo-hakari: 
> For hakari to work correctly, you must add your Cargo.lock to version control, even if you don't have any binary crates. This is because patch version bumps in dependencies can add or remove features or even entire transitive dependencies.

I don't think we have `Cargo.lock` in our repo?

tsunrise (changes requested):

manta-workspace-hack/Cargo.toml
bhgomes: Hmm, interesting. Good catch, I hadn't seen this. Not sure why this is mandatory though let me think about it.

bhgomes: I don't actually think this is needed. The `Cargo.lock` for these crates should be represented by the "latest possible Cargo.lock" so if you were to delete `Cargo.lock` and `target` (which is what the CI does) then call `cargo update` (gets a new `Cargo.lock` file) this is what the latest one should be. If `cargo-hakari` on the CI disagrees with he `Cargo.lock` that a dev used, then the dev should update their `Cargo.lock` and re-run `cargo hakari`. I don't see this to be an issue in practice since all of us are pretty up-to-date, this would happen with a dev that hasn't worked on the project in a while and has an older `Cargo.lock`.

tsunrise: Sounds good! Let's ignore this for now

tsunrise (approved):

BoyuanFeng (approved):

PR #176, opened 2022-07-18
feat: upgrade manta-crypto abstractions for new circuits

bhgomes:
- Upgrade `Derived` allocation mode to arbitrary type flags
- Split out the accumulator model trait into a `Types` trait
- Add `ItemHashFunction` for explicit "leaf hash" constructions for general accumulators (with `ItemHashAccumulatorModel` and `ItemHashAccumulator` wrappers)
- Move allocation of `MembershipProof` to `Derived` mode
- Add `assert_equal` to `PartialEq` to lift efficient assertion implementations to the `AssertEq` trait
- Add `NonNative` marker trait to allow distinguishing `()` from non-native compilers (this allows `(): Has<T>`)
- Add `Zero` and `One` traits
- Add more variable allocation and `PartialEq` implementations for compound types
- Add helper alias types for encryption
- Add `CheckedAdd` and `CheckedSub` traits to numeric utilities
- Add `all_unequal` utility function

---

### DIFF
--- a/manta-crypto/src/eclair/cmp.rs
+++ b/manta-crypto/src/eclair/cmp.rs
@@ -16,10 +16,15 @@
 
 //! Comparison
 
-use crate::eclair::{bool::Bool, ops::Not, Has};
+use crate::eclair::{
+    bool::{Assert, Bool},
+    ops::Not,
+    Has,
+};
+use core::cmp;
 
 /// Partial Equivalence Relations
-pub trait PartialEq<Rhs, COM>
+pub trait PartialEq<Rhs, COM = ()>
 where
     Rhs: ?Sized,
     COM: Has<bool> + ?Sized,
@@ -35,9 +40,23 @@ where
     {
         self.eq(other, compiler).not(compiler)
     }
+
+    /// Asserts that `self` and `rhs` are equal.
+    ///
+    /// # Implementation Note
+    ///
+    /// This method is an optimization path for the case when comparing for equality and then
+    /// asserting is more expensive than a custom assertion.
+    #[inline]
+    fn assert_equal(&self, rhs: &Rhs, compiler: &mut COM)
+    where
+        COM: Assert,
+    {
+        let are_equal = self.eq(rhs, compiler);
+        compiler.assert(&are_equal);
+    }
 }
 
-/* FIXME: We cannot implement this yet.
 impl<T, Rhs> PartialEq<Rhs> for T
 where
     T: cmp::PartialEq<Rhs>,
@@ -52,15 +71,12 @@ where
         self.ne(rhs)
     }
 }
-*/
 
 /// Equality
-pub trait Eq<COM>: PartialEq<Self, COM>
+pub trait Eq<COM = ()>: PartialEq<Self, COM>
 where
     COM: Has<bool>,
 {
 }
 
-/* FIXME: We cannot implement this yet.
 impl<T> Eq for T where T: cmp::Eq {}
-*/
--- a/manta-crypto/src/eclair/cmp.rs
+++ b/manta-crypto/src/eclair/cmp.rs
@@ -21,7 +21,6 @@ use crate::eclair::{
     ops::Not,
     Has,
 };
-use core::cmp;
 
 /// Partial Equivalence Relations
 pub trait PartialEq<Rhs, COM = ()>
@@ -57,6 +56,7 @@ where
     }
 }
 
+/* TODO:
 impl<T, Rhs> PartialEq<Rhs> for T
 where
     T: cmp::PartialEq<Rhs>,
@@ -71,6 +71,7 @@ where
         self.ne(rhs)
     }
 }
+*/
 
 /// Equality
 pub trait Eq<COM = ()>: PartialEq<Self, COM>
@@ -79,4 +80,6 @@ where
 {
 }
 
+/* TODO:
 impl<T> Eq for T where T: cmp::Eq {}
+*/
--- a/manta-crypto/src/merkle_tree/tree.rs
+++ b/manta-crypto/src/merkle_tree/tree.rs
@@ -241,7 +241,7 @@ where
         L: IntoIterator<Item = &'l Leaf<C>>,
     {
         let mut tree = Self::new(parameters);
-        tree.extend(parameters, leaves).then(move || tree)
+        tree.extend(parameters, leaves).then_some(tree)
     }
 
     /// Builds a new merkle tree with the given `leaves` returning `None` if the slice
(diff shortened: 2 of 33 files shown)


SupremoUGH:
Two comments: 
1) About assert_equal for PartialEq
I don't see how calling the new assert_equal for AssertEq which calls an assert_equal defined for PartialEq, which in turn is the old assert_equal for AssertEq can be more efficient than directly calling the old one. Why is that?
2) About NonNative: is the idea the do a general splitting everywhere of those implementations that depend on COM into implementations that don't and implementations that depend on NonNative COM (i.e. is the ultimate goal to generalize the splitting you've done when implementing accumulator::Model for Parameters?). I see why that might be interesting (code for () can be much simpler than code for a generic compiler COM), but in the particular case you apply the splitting it doesn't seem so, as both implementations look almost identical to each other and also to the general code for any COM that existed before.

manta-pay/src/crypto/constraint/arkworks/mod.rs
```diff
@@ -470,6 +468,19 @@ where
     }
 }
 
+impl<F> BitAnd<Self, R1CS<F>> for Boolean<F>
```
SupremoUGH, line 469: I don't understand this. What is the difference between this implementation, which defines a function bitand for Boolean<F>, and the already existing function and? I guess you want to use the fact that Boolean<F> implements the trait BitAnd for some abstractions later on, otherwise I don't see the point of this implementation.

manta-crypto/src/eclair/cmp.rs
```diff
@@ -35,9 +40,23 @@ where
...
+    /// This method is an optimization path for the case when comparing for equality and then
+    /// asserting is more expensive than a custom assertion.
+    #[inline]
+    fn assert_equal(&self, rhs: &Rhs, compiler: &mut COM)
+    where
+        COM: Assert,
+    {
+        let are_equal = self.eq(rhs, compiler);
+        compiler.assert(&are_equal);
+    }
```
SupremoUGH, line 56: I don't see how calling the new assert_equal for AssertEq which calls an assert_equal defined for PartialEq, which in turn is the old assert_equal for AssertEq can be more efficient than directly calling the old one. Why is that?

manta-crypto/src/merkle_tree/tree.rs
```diff
@@ -692,16 +699,25 @@ where
...
     type Output = Root<C, COM>;
+}
+
+impl<C, COM> accumulator::Model<COM> for Parameters<C, COM>
+where
+    C: Configuration<COM> + ?Sized,
+    COM: Has<bool> + NonNative,
+    InnerDigest<C, COM>: ConditionalSwap<COM> + constraint::PartialEq<InnerDigest<C, COM>, COM>,
+    LeafDigest<C, COM>: ConditionalSwap<COM>,
+{
```
SupremoUGH, line 720: About the NonNative trait: is the idea to do a general splitting everywhere of those implementations that depend on COM into implementations that don't and implementations that depend on NonNative COM (i.e. is the ultimate goal to generalize the splitting you've done when implementing accumulator::Model for Parameters?). I see why that might be interesting (code for () can be much simpler than code for a generic compiler COM), but in the particular case you apply the splitting it doesn't seem so, as both implementations look almost identical to each other and also to the general code for any COM that existed before.

SupremoUGH (changes requested):

manta-pay/src/crypto/constraint/arkworks/mod.rs
bhgomes: This trait is used to define generic circuits which support an `&` operation (called `BitAnd` in Rust). This implementation allows us to write circuits in a generic way without relying on the arkworks backend circuit definition system.

manta-crypto/src/eclair/cmp.rs
bhgomes: Here we are allowing the specific type to customize the behavior of `assert_eq` whenever it can be more efficient than the default behavior. If we left this implementation to the compiler, it would have to provide an explicit implementation that worked over all types and would have to be generic over all of them. Instead, we let the type handle this behavior when it knows how to do this better. Here's an example:
```rust
struct Struct<A, B> {
    a: A,
    b: B,
}

impl<A, B, COM> PartialEq<Self, COM> for Struct<A, B>
where
    A: PartialEq<Self, COM>,
    B: PartialEq<Self, COM>,
{
    fn assert_equal(&self, rhs: &Self, compiler: &mut COM) {
        compiler.assert_eq(self.a, rhs.a);
        compiler.assert_eq(self.b, rhs.b);
    }
}
```
instead of
```rust
fn assert_equal(&self, rhs: &Self, compiler: &mut COM) {
    let are_equal = self.a.eq(rhs.a, compiler).bitand(self.b.eq(rhs.b, compiler), compiler);
    compiler.assert(are_equal);
}
```
Notice that in this second one we have to construct the intermediate boolean `are_equal` and we have to pay for a `bitand` operation as well.

manta-crypto/src/merkle_tree/tree.rs
bhgomes: The `NonNative` trait used to be "implemented" by the `Has<bool>` trait when `()` did not implement it. It now does, so we cannot use that as the way of distinguishing Rust from non-Rust compilers. Instead, the `NonNative` trait allows us to explicitly distinguish them so we can give different implementations for each case. For merkle tree path verification, we actually have two different data structures `Path` and `PathVar` which are not the same and so we need two different verification functions.

SupremoUGH (dismissed):

manta-util/src/num.rs
```diff
@@ -49,3 +49,21 @@ impl Ceil<usize> for f64 {
         self.ceil() as usize
     }
 }
+
+/// Checked Addition
+pub trait CheckedAdd<Rhs = Self> {
+    /// Output Type
+    type Output;
+
+    /// Checked integer subtraction. Computes `self + rhs`, returning `None` if overflow occurred.
```
BoyuanFeng, line 58:
```suggestion
    /// Checked integer addition. Computes `self + rhs`, returning `None` if overflow occurred.
```

BoyuanFeng (changes requested):
Thanks for this great PR! Two minor comments:

manta-util/src/num.rs
bhgomes: Good catch!

SupremoUGH (approved):
Everything seems alright since the last review. Just one question to see if I understand correctly what's going on in the last commit. Is it that you are now required to explicitly mention the generic lifetime (even if you don't actually use it, hence the underscore) because of the new nightly released, and no actual change to the code happened?

BoyuanFeng (approved):

PR #177, opened 2022-07-21
[DO NOT MERGE] feat: ceremony for trusted setup

tsunrise:
Add ceremony implementations for trusted setup. 
Reference: https://github.com/Manta-Network/trusted-setup/blob/server-client-features/

**This PR will be split to several smaller PRs for review.**
- Queue, Registry, and Signature #213 
- Client and Server Primitives #238 
- Client and Server Binaries (PR not made yet)


manta-trusted-setup/src/ceremony/signature.rs
```diff
@@ -0,0 +1,250 @@
...
+            )
+            .expect("Failed to decode keypair from bytes"); // todo: error handling
+
+            let msg = domain_tag
+                .iter()
+                .chain(self.0.iter())
+                .copied()
+                .collect::<Vec<_>>();
+
+            let sig = keypair.sign(&msg);
```
BoyuanFeng, line 190: It seems that we only sign `domain_tag` here instead of a general message `msg`. Should we explicitly split the concept of `domain_tag` and `msg` here? For example, `domain_tag` should be some `u8` value while `msg` is an array of `u8`.

BoyuanFeng (changes requested):

manta-trusted-setup/src/ceremony/signature.rs
```diff
@@ -0,0 +1,250 @@
...
+{
+    /// Signature specific to this message
+    type Signature: Sized;
+    /// Sign the message
+    fn sign(
+        &self,
+        domain_tag: &S::DomainTag,
+        public_key: &S::PublicKey,
+        private_key: &S::PrivateKey,
+    ) -> Result<Self::Signature, CeremonyError>;
```
BoyuanFeng, line 71: in my understanding, we only need the private key to sign a message. Why should we use both private key and public key here?

manta-trusted-setup/src/ceremony/signature.rs
```diff
@@ -0,0 +1,250 @@
...
+        domain_tag: &S::DomainTag,
+        public_key: &S::PublicKey,
+    ) -> Result<(), CeremonyError> {
+        self.message
+            .verify_integrity(domain_tag, public_key, &self.signature)
+    }
+}
+
+/// Implementation of Signature Scheme using Ed-Dalek
+pub mod ed_dalek {
```
BoyuanFeng, line 105: We have a schnorr signature in `manta-crypto/src/signature.rs`. Should we use that?

BoyuanFeng (changes requested):

manta-trusted-setup/src/ceremony/requests.rs
```diff
@@ -0,0 +1,260 @@
...
+// {
+//     #[inline]
+//     fn sign(
+//         &self,
+//         nounce: &S::Nounce,
+//         public_key: &S::PublicKey,
+//         private_key: &S::PrivateKey,
+//     ) -> Result<S::Signature, CeremonyError> {
+//         // sign the public key
+//         public_key.sign(nounce, public_key, private_key)
```
BoyuanFeng, line 114: Should we sign both the public key and other information such as identifier and priority?

BoyuanFeng (changes requested):

manta-trusted-setup/src/ceremony/signature.rs
```diff
@@ -0,0 +1,226 @@
...
+        ) -> Result<Self::Signature, CeremonyError>
+        where
+            M: ?Sized + AsRef<[u8]>,
+        {
+            let mut message_concatenated =
+                bincode::serialize(nonce).expect("Serializing nonce should not fail");
+
+            message_concatenated.extend_from_slice(message.as_ref());
+            Ok(Signature(into_array_unchecked(
+                Keypair::from_bytes(&[&private_key.0[..], &public_key.0[..]].concat())
```
BoyuanFeng, line 175: Take a seed phrase and generate keypair from the seed

manta-trusted-setup/src/ceremony/coordinator.rs
```diff
@@ -58,84 +63,79 @@ where
...
-            .expect("participant in the queue should be registered"))
+    /// Pops the current contributor and moves to the back.
+    #[inline]
+    pub fn skip_current_contributor(&mut self) -> Result<(), CeremonyError> {
+        let (priority, identifier) = self
+            .queue
+            .pop()
+            .ok_or(CeremonyError::WaitingQueueEmpty)
+            .expect("Poping the current participant should succeed.");
+        self.queue.push(priority, identifier);
```
BoyuanFeng, line 138: What if there is only 1 person left in the queue but he never contributes? Then this is a infinite loop.

manta-trusted-setup/src/bin/groth16_phase2_client.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/// Prompts the client information.
+#[inline]
+pub fn prompt_client_info() -> Result<(PublicKey<C>, PrivateKey<C>), Error> {
+    println!(
+        "Please enter your {} that you get when you registered yourself using this tool.",
+        "Secret".italic()
+    );
+    let secret_str: String = Input::with_theme(&ColorfulTheme::default())
```
tsunrise, line 110: @BoyuanFeng You may want to use `Password` instead of `Input` for secret: https://docs.rs/dialoguer/0.10.2/dialoguer/struct.Password.html

manta-trusted-setup/src/bin/groth16_phase2_client.rs
```diff
@@ -0,0 +1,279 @@
...
+                return Err(Error::UnexpectedError(
+                    "unexpected error when query mpc state since finding a bad request."
+                        .to_string(),
+                ))
+            }
+            Err(CeremonyError::AlreadyContributed) => return Err(Error::AlreadyContributed),
+            Ok(message) => match message {
+                QueryResponse::QueuePosition(position) => {
+                    println!("Your current position is {}.\n", position);
+                    continue;
```
tsunrise, line 220: We may also need to add wait time?

manta-trusted-setup/src/ceremony/coordinator.rs
```diff
@@ -150,41 +123,63 @@ where
...
     }
 
-    /// Updates the MPC state and challenge using client's contribution. If the contribution is valid,
-    /// the participant will be removed from the waiting queue, and cannot participate in this ceremony
-    /// again.
+    /// Updates the expired lock by reducing the priority of it's participant and setting it's
+    /// contained value to the new front of the queue. The previous participant in the lock is
+    /// returned.
+    #[inline]
+    fn update_expired_lock(&mut self) -> Option<ParticipantIdentifier<C>> {
```
tsunrise, line 130: @bhgomes I am not sure if `lock` is the best name to describe `Timed<Option<ParticipantIdentifier>>`. This is because `Coordinator` itself is also guarded by a "lock" which has different meaning.

I don't mind if we make the name more verbose, something like `current_paritipant`

bhgomes: @tsunrise yeah lets call it `participant_lock` 👍

bhgomes:
TODO: Review security requirements here: https://github.com/dalek-cryptography/ed25519-dalek

bhgomes:
Closed in favor of #259 and #257.

PR #193, opened 2022-08-02
Bn254 Backend for Perpetual Powers of Tau

### LINKED ISSUE #174: BN254 Trusted Setup Backend
Since we are most likely switching curves, we need to look into the available trusted setups for phase1 that we can integrate into our backend.

GhostOfGauss:
PPoT Phase 1 for Bn254
---
Allows us to use the PPoT Phase 1 for Bn254.

Closes #174 


manta-trusted-setup/Cargo.toml
```diff
@@ -37,6 +37,8 @@ test = ["manta-crypto/test"]
 [dependencies]
 ark-groth16 = { version = "0.3.0", default-features = false }
 ark-poly = { version = "0.3.0", default-features = false }
+ark-relations = { version = "0.3.0", default-features = false }
+ark-serialize = { version = "0.3.0", default-features = false, features = ["derive", "std"] }
```
GhostOfGauss, line 41: Is the `std` feature okay here ?

manta-trusted-setup/src/groth16/bn254/manta_pay.rs
```diff
@@ -0,0 +1,363 @@
...
+
+//     fn serialize_compressed<W>(item: &short_weierstrass_jacobian::GroupAffine<P>, writer: &mut W) -> Result<(), std::io::Error>
+//     where
+//         W: Write {
+//         todo!()
+//     }
+
+//     fn compressed_size(item: &short_weierstrass_jacobian::GroupAffine<P>) -> usize {
+//         todo!()
+//     }
```
GhostOfGauss, line 205: Here I just want to use its `CanonicalSerialize` implementation but I can't figure out how to convert errors

manta-trusted-setup/src/groth16/bn254/manta_pay.rs
```diff
@@ -0,0 +1,363 @@
...
+    Reclaim::unknown_constraints(FullParameters::new(
+        &parameters,
+        <MerkleForest<_, _> as Accumulator>::model(&utxo_accumulator),
+    ))
+}
+
+/// Generates our `Reclaim` circuit with known variables
+pub fn reclaim_circuit_known() -> R1CS<Fr> {
+    todo!()
+}
```
GhostOfGauss, line 233: (Currently reworking this)

manta-trusted-setup/src/groth16/bn254/manta_pay.rs
```diff
@@ -0,0 +1,363 @@
...
+        .expect("unable to create parameter file in this directory");
+    CanonicalSerialize::serialize_uncompressed(&pk, &mut file).unwrap();
+
+    // Check that this was written correctly:
+    let mut reader = OpenOptions::new()
+        .read(true)
+        .open("phase2_reclaim_pk")
+        .expect("file not found");
+    let pk_read: ProvingKey<Bn254> =
+        CanonicalDeserialize::deserialize_uncompressed(&mut reader).unwrap();
```
GhostOfGauss, line 273: This can be `deserialize_unchecked`

manta-trusted-setup/src/groth16/bn254/manta_pay.rs
```diff
@@ -0,0 +1,363 @@
...
+        transcript.rounds,
+    )
+    .expect("Verifying all transformations failed.");
+    println!("Verified phase 2 contributions in {:?}", now.elapsed());
+
+    // Generate and verify a proof
+    println!("Generating a proof");
+    let now = Instant::now();
+    let mut rng = ChaCha20Rng::from_seed([0; 32]);
+    let mut utxo_accumulator = UtxoAccumulator::new(rng.gen());
```
GhostOfGauss, line 343: (currently reworking this)

manta-trusted-setup/src/groth16/bn254/ppot.rs
```diff
@@ -0,0 +1,525 @@
...
+            .expect("Consuming ratio proof of beta failed.");
+        into_array_unchecked(hasher.0.finalize())
+    }
+}
+
+// impl Serializer<<Self as Pairing>::G1> for PpotCeremony {
+//     fn serialize_unchecked<W>(item: &<Self as Pairing>::G1, writer: &mut W) -> Result<(), std::io::Error>
+//     where
+//         W: ark_serialize::Write {
+// TODO : How do you distinguish between G1 and G2? They both have type sw_jacobian::GroupAffine<P>
```
GhostOfGauss, line 127: Actually there's no need to impl Serializer for PPoT.  But same question holds for Deserializer: how do I account for the fact that deserialization works differently for G1 and G2 points when these share the GroupAffine<P> type?

manta-trusted-setup/src/groth16/kzg.rs
```diff
@@ -384,7 +384,7 @@ where
 }
 
 /// Contribution Accumulator
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(CanonicalDeserialize, CanonicalSerialize, Clone, Debug, Eq, Hash, PartialEq)]
```
GhostOfGauss, line 387: I commented out the `impl CanonicalSerialize for Accumulator` in this file because I was having trouble with `Serializer` for the PPoT configuration

manta-pay/Cargo.toml
```diff
@@ -40,7 +40,10 @@ required-features = ["clap", "groth16", "simulation"]
 # Enable Arkworks Backend
 arkworks = [
     "ark-bls12-381",
+    "ark-ec",
     "ark-ed-on-bls12-381",
+    "ark-ff",
+    "ark-serialize",
```
bhgomes, line 46: These are accessed through `manta_crypto::arkworks::ec` or `manta_crypto::arkworks::ff`, etc. and don't need to be imported anymore.

manta-pay/Cargo.toml
```diff
@@ -101,8 +104,11 @@ websocket = [
 [dependencies]
 aes-gcm = { version = "0.9.4", default-features = false, features = ["aes", "alloc"] }
 ark-bls12-381 = { version = "0.3.0", optional = true, default-features = false, features = ["curve"] }
+ark-ec = { version = "0.3.0", optional = true, default-features = false }
 ark-ed-on-bls12-381 = { version = "0.3.0", optional = true, default-features = false, features = ["r1cs"] }
+ark-ff = { version = "0.3.0", optional = true, default-features = false }
 ark-groth16 = { version = "0.3.0", optional = true, default-features = false }
+ark-serialize = { version = "0.3.0", optional = true, default-features = false }
```
bhgomes, line 111: See above comment.

manta-trusted-setup/src/groth16/bn254/manta_pay.rs
```diff
@@ -0,0 +1,239 @@
...
+use manta_util::into_array_unchecked;
+
+/// Configuration for a Phase1 Ceremony large enough to support MantaPay circuits
+#[derive(CanonicalDeserialize, CanonicalSerialize, Debug, PartialEq, Eq)]
+pub struct MantaPaySetupCeremony;
+
+impl Size for MantaPaySetupCeremony {
+    const G1_POWERS: usize = (Self::G2_POWERS << 1) - 1;
+
+    const G2_POWERS: usize = 1 << 19;
```
bhgomes, line 48:
```suggestion
    const G1_POWERS: usize = (Self::G2_POWERS << 1) - 1;
    const G2_POWERS: usize = 1 << 19;
```

manta-trusted-setup/src/groth16/bn254/manta_pay.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    type Pairing = Bn254;
+
+    fn g1_prime_subgroup_generator() -> Self::G1 {
+        G1Affine::prime_subgroup_generator()
+    }
+
+    fn g2_prime_subgroup_generator() -> Self::G2 {
+        G2Affine::prime_subgroup_generator()
+    }
```
bhgomes, line 70:
```suggestion
    type Scalar = Fr;
    type G1 = G1Affine;
    type G1Prepared = <Self::Pairing as PairingEngine>::G1Prepared;
    type G2 = G2Affine;
    type G2Prepared = <Self::Pairing as PairingEngine>::G2Prepared;
    type Pairing = Bn254;
    
    #[inline]
    fn g1_prime_subgroup_generator() -> Self::G1 {
        G1Affine::prime_subgroup_generator()
    }

    #[inline]
    fn g2_prime_subgroup_generator() -> Self::G2 {
        G2Affine::prime_subgroup_generator()
    }
```

manta-trusted-setup/src/groth16/bn254/manta_pay.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    fn serialize_uncompressed<W>(
+        item: &short_weierstrass_jacobian::GroupAffine<P>,
+        writer: &mut W,
+    ) -> Result<(), io::Error>
+    where
+        W: Write,
+    {
+        CanonicalSerialize::serialize_uncompressed(item, writer)
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))
```
bhgomes, line 193: Factor out this error function:

```rust
#[inline]
fn from_serialization_error(err: SerializationError) -> io::Error {
    io::Error::new(io::ErrorKind::Other, err)
}
```

something like this.

manta-trusted-setup/src/groth16/bn254/manta_pay.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    fn deserialize_compressed<R>(
+        reader: &mut R,
+    ) -> Result<short_weierstrass_jacobian::GroupAffine<P>, Self::Error>
+    where
+        R: Read,
+    {
+        CanonicalDeserialize::deserialize_uncompressed(reader)
+    }
```
bhgomes, line 238: Mark all these serialization functions as `#[inline]` (especially because they are all a single line).

manta-trusted-setup/src/groth16/bn254/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Bn254 Backend for MantaPay Groth16 Trusted Setup
+
+/// Configuration for MantaPay circuits
+#[cfg(test)]
+pub mod manta_pay;
+/// Utilities to use the accumulator of the Perpetual Powers of Tau ceremony
+#[cfg(test)]
+pub mod ppot;
```
bhgomes, line 24:
```suggestion
#[cfg(test)]
pub mod manta_pay;

#[cfg(test)]
pub mod ppot;
```

bhgomes: Any docstrings should appear as `//!` docs in the module itself.

manta-trusted-setup/src/groth16/mod.rs
```diff
@@ -19,5 +19,7 @@
 pub mod kzg;
 pub mod mpc;
 
+#[cfg(test)]
+pub mod bn254;
 #[cfg(test)]
 pub mod test;
```
bhgomes, line 25:
```suggestion
#[cfg(test)]
pub mod bn254;

#[cfg(test)]
pub mod test;
```

manta-trusted-setup/src/groth16/mpc.rs
```diff
@@ -135,6 +135,27 @@ pub fn specialize_to_phase_2<G1, G2>(
         );
 }
 
+/// Adds dummy `z_i * 0 = 0` constraint for each public input `z_i`.  This ensures
+/// non-malleability of Groth16 proofs even if some public inputs are otherwise unconstrained.
+#[inline]
+pub fn dummy_constraints<G>(
```
bhgomes, line 141: Should include some verb in the name like `add_dummy_constraints`.

manta-trusted-setup/src/groth16/test.rs
```diff
@@ -63,10 +63,10 @@ impl HasDistribution for Test {
 impl Pairing for Test {
     type Scalar = Fr;
     type G1 = G1Affine;
-    type G1Prepared = <Bls12_381 as PairingEngine>::G1Prepared;
+    type G1Prepared = <Bn254 as PairingEngine>::G1Prepared;
     type G2 = G2Affine;
-    type G2Prepared = <Bls12_381 as PairingEngine>::G2Prepared;
-    type Pairing = Bls12_381;
+    type G2Prepared = <Bn254 as PairingEngine>::G2Prepared;
+    type Pairing = Bn254;
```
bhgomes, line 69: Like above use `Self::Pairing` as the argument in `<Self::Pairing as PairingEngine>::G2Prepared`

manta-trusted-setup/src/groth16/bn254/ppot.rs
```diff
@@ -0,0 +1,1102 @@
...
+#[cfg(feature = "rayon")]
+use manta_util::rayon::prelude::ParallelIterator;
+
+/// Configuration of the Perpetual Powers of Tau ceremony
+pub struct PpotCeremony;
+
+impl Size for PpotCeremony {
+    const G1_POWERS: usize = (Self::G2_POWERS << 1) - 1;
+
+    const G2_POWERS: usize = 1 << 28;
```
bhgomes, line 52:
```suggestion
    const G1_POWERS: usize = (Self::G2_POWERS << 1) - 1;
    const G2_POWERS: usize = 1 << 28;
```

manta-trusted-setup/src/groth16/bn254/ppot.rs
```diff
@@ -0,0 +1,1102 @@
...
+
+    type Pairing = Bn254;
+
+    fn g1_prime_subgroup_generator() -> Self::G1 {
+        G1Affine::prime_subgroup_generator()
+    }
+
+    fn g2_prime_subgroup_generator() -> Self::G2 {
+        G2Affine::prime_subgroup_generator()
+    }
```
bhgomes, line 74:
```suggestion
    type Scalar = Fr;
    type G1 = G1Affine;
    type G1Prepared = <Self::Pairing as PairingEngine>::G1Prepared;
    type G2 = G2Affine;
    type G2Prepared = <Self::Pairing as PairingEngine>::G2Prepared;
    type Pairing = Bn254;

    #[inline]
    fn g1_prime_subgroup_generator() -> Self::G1 {
        G1Affine::prime_subgroup_generator()
    }

    #[inline]
    fn g2_prime_subgroup_generator() -> Self::G2 {
        G2Affine::prime_subgroup_generator()
    }
```

manta-trusted-setup/src/groth16/bn254/ppot.rs
```diff
@@ -0,0 +1,1102 @@
...
+    pairing::Pairing,
+};
+use manta_util::{cfg_iter, into_array_unchecked, vec::Vec};
+use memmap::Mmap;
+
+#[cfg(feature = "rayon")]
+use manta_util::rayon::prelude::ParallelIterator;
+
+/// Configuration of the Perpetual Powers of Tau ceremony
+pub struct PpotCeremony;
```
bhgomes, line 47: How much code is shared between this configuration and `MantaPaySetupCeremony`? If the manta pay system is just a subaccumulator we should make this explicit somehow and inherit as much as we can from the ceremony code.

manta-pay/src/crypto/constraint/arkworks/pairing.rs
```diff
@@ -175,3 +175,163 @@ pub mod bls12 {
         type Deserialize = G2Prepared<P>;
     }
 }
+
+/// Bn254 Utilities
```
bhgomes, line 179: How different is this from the BLS one? If it's the same, please write a macro that generates this code for an arbitrary model and then instantiate it for each one.

bhgomes (changes requested):
Also seems that `phase2_reclaim_pk` is a stray file.

manta-trusted-setup/src/groth16/bn254/ppot.rs
GhostOfGauss: Their implementations of `Pairing` and `groth16::kzg::Configuration` are identical. Is a macro the best way to do this?

bhgomes: If the only difference is the size then you can do something like this:

```rust
pub struct PerpetualPowersOfTauCeremony<const POWERS: usize>;
```

and then give the implementations for this type instead.

GhostOfGauss: That works well for implementations of `Pairing` and `groth16::kzg::Configuration` but not for `Serializer` and `Deserializer` because we want to use arkworks canonical serialization for the `MantaPaySetupCeremony` but bellman-style serialization for the original ppot ceremony.  Any suggestion there?

bhgomes: When in doubt, add more paramters!

```rust
pub struct PerpetualPowersOfTauCeremony<S, const POWERS: usize>;
```

where `S` implements the appropriate `Serializer` and `Deserializer`.

manta-trusted-setup/src/groth16/test/bn254/ppot.rs
```diff
@@ -0,0 +1,1125 @@
...
+use memmap::Mmap;
+
+#[cfg(feature = "rayon")]
+use manta_util::rayon::prelude::ParallelIterator;
+
+#[derive(Debug, PartialEq, Eq)]
+/// Configuration of the Perpetual Powers of Tau ceremony
+pub struct PerpetualPowersOfTauCeremony<S, const POWERS: usize> {
+    __: PhantomData<S>,
+}
```
bhgomes, line 50: A few things to keep in mind:

1. Put the docstring on the first line.
2. Because the Rust `derive` macro generates the wrong semantics, use `derivative` and since this is just a `PhantomData` marker struct, implement all the standard traits (see below) in alphabetical order.
3. Since it's only a marker struct, use the tuple-struct form of the constructor `Struct(_)` rather than `Struct { .. }`.

```suggestion
/// Configuration of the Perpetual Powers of Tau Ceremony
#[derive(derivative::Derivative)]
#[derivative(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
pub struct PerpetualPowersOfTauCeremony<S, const POWERS: usize>(PhantomData<S>);
```

manta-trusted-setup/src/groth16/test/bn254/ppot.rs
```diff
@@ -0,0 +1,1125 @@
...
+        S::deserialize_unchecked(reader)
+    }
+
+    fn deserialize_compressed<R>(reader: &mut R) -> Result<T, Self::Error>
+    where
+        R: Read,
+    {
+        S::deserialize_compressed(reader)
+    }
+}
```
bhgomes, line 143: For this `trait`, since `Self` is wrapping an implementation, you'll need to forward _all_ the methods not just the required ones to have the correct semantics. So make sure to also implement `check` and `deserialize_uncompressed`. Also, make sure to implement them in the same order as written in the trait and mark all the methods as `#[inline]` (these 100% should be marked).

manta-trusted-setup/src/groth16/test/bn254/ppot.rs
```diff
@@ -0,0 +1,1125 @@
...
+    }
+
+    fn compressed_size(item: &T) -> usize {
+        S::compressed_size(item)
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+/// (De)Serialization used in the original PPoT ceremony
+pub struct PpotSerializer {}
```
bhgomes, line 181: Same structuring comment as above, and also use `pub struct PpotSerializer;`.

bhgomes (changes requested):

manta-trusted-setup/src/groth16/ppot/hashing.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        let _prev = Accumulator::<SubCeremony>::verify_transform(
+            prev_accumulator,
+            next_accumulator,
+            challenge_hash,
+            proof.cast_to_subceremony(),
+        )
+        .unwrap();
+    }
+}
```
SupremoUGH, line 226: Q: If you take the files from an external source, and considering that the hash takes a challenge which will, in principle, depend on the ceremony size, when we take the subceremony we have no control over the new hashes, right? So even if the transition from BIG_FILE_1 to BIG_FILE_2 satisfies the test, when we reduce it to SMALL_FILE_1 and SMALL_FILE_2, shouldn't it fail?

manta-trusted-setup/src/groth16/ppot/kzg.rs
```diff
@@ -0,0 +1,199 @@
...
+
+/// Configuration of the Perpetual Powers of Tau Ceremony
+#[derive(derivative::Derivative)]
+#[derivative(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct PerpetualPowersOfTauCeremony<S, const POWERS: usize>(PhantomData<S>);
+
+impl<S, const POWERS: usize> Size for PerpetualPowersOfTauCeremony<S, POWERS> {
+    const G1_POWERS: usize = (Self::G2_POWERS << 1) - 1;
+    const G2_POWERS: usize = POWERS;
+}
```
SupremoUGH, line 45: Why is this? According to the Groth16 paper, we should have the same number of powers for both EC.

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,964 @@
...
+    Ok(powers)
+}
+
+/// Reads `size` many elements of `element_type` from PPoT challenge file.
+#[inline]
+pub fn read_g2_powers<S>(
+    readable_map: &Mmap,
+    element: ElementType,
+    compression: Compressed,
+) -> Result<Vec<GroupAffine<<Parameters as BnParameters>::G2Parameters>>, PointDeserializeError>
```
SupremoUGH, line 626: Why not G2Affine?

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,964 @@
...
+                panic!() // This should actually just return the NotInSubgroup error
+            }
+        }),
+    }
+    Ok(powers)
+}
+
+/// Reads `size` many elements of `element_type` from PPoT challenge file.
+#[inline]
+pub fn read_g2_powers<S>(
```
SupremoUGH, line 622: Why not a macro that generates both `read_g1_powers` and `read_g2_powers`?

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,964 @@
...
+        + (PpotCeremony::G1_POWERS + 2 * PpotCeremony::G2_POWERS) * 32
+        + (PpotCeremony::G2_POWERS + 1) * 64;
+    let reader = readable_map
+        .get(position..position + 6 * 64 + 3 * 128)
+        .expect("cannot read point data from file");
+
+    <Proof<_> as CanonicalDeserialize>::deserialize_uncompressed(reader)
+}
+
+/// Extracts a subaccumulator of size `required_powers`. Specific to PPoT challenge file.
```
SupremoUGH, line 687: `required_powers` is not anywhere in the function

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,964 @@
...
+                .map(&reader)
+                .expect("unable to create a memory map for input")
+        };
+        let response_acc =
+            read_subaccumulator::<SubCeremony>(&response_map, Compressed::Yes).unwrap();
+        println!("Read compressed accumulator in {:?}", now.elapsed());
+
+        assert_eq!(challenge_acc, response_acc)
+    }
+}
```
SupremoUGH, line 964: Would it make sense to also run some tests against known (de)serialization implementations?

SupremoUGH (changes requested):

manta-trusted-setup/src/groth16/mpc.rs
```diff
@@ -135,6 +134,24 @@ pub fn specialize_to_phase_2<G1, G2>(
         );
 }
 
+/// Adds dummy `z_i * 0 = 0` constraint for each public input `z_i`.  This ensures non-malleability
+/// of Groth16 proofs even if some public inputs are otherwise unconstrained.
```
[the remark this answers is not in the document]
GhostOfGauss: Yeah this is a weird technical point.  It's related to a hypothetical situation where you might want to allow a circuit to have public inputs that are not part of any constraint.  I don't know offhand what situation that might be, but having unconstrained public inputs leads to "malleability" of the proofs, which is undesirable.  It's explained better [here](https://geometry.xyz/notebook/groth16-malleability). 

I'm unsure how much explanation to give it in the documentation.  Every Groth16 implementation seems to do this and no one does a good job of documenting it.

manta-trusted-setup/src/groth16/ppot/hashing.rs
```diff
@@ -0,0 +1,226 @@
...
+    pub domain_tag: u8,
+}
+
+impl<S, const POWERS: usize, const N: usize>
+    HashToGroup<PerpetualPowersOfTauCeremony<S, POWERS>, [u8; N]> for PpotHasher
+where
+    S: Serializer<G1Affine, G1>,
+{
+    #[inline]
+    fn hash(&self, challenge: &[u8; N], ratio: (&G1Affine, &G1Affine)) -> G2Affine {
```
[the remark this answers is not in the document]
GhostOfGauss: I agree, `pair` is a better name. I probably called it `ratio` because I think of that pair as representing a ratio, but it's probably clearer to say `pair`.

manta-trusted-setup/src/groth16/ppot/hashing.rs
GhostOfGauss: Yep, that's right. The proof is tied to the hash of the file, so if you change the file you change the proof and nothing works. So what we do is we always hash the _full file_ even if we're only verifying a subset of the accumulator it contains. This is why we had to download all 10 Tb of data to verify PPoT even though we only use 1/500th of the actual accumulators.

In this test we use a shortcut: if you look [here](https://github.com/Manta-Network/manta-rs/blob/3dbf4b5faec4d4a0c7f7c83d4dcd01528fbd8589/manta-trusted-setup/src/groth16/ppot/hashing.rs#L211) you'll see that instead of hashing a file we actually read the hash from another file.  That's because `response_k` is _supposed_ to contain the hash of `challenge_k` at the beginning of the file.  Now it might not, but if you optimistically assume that it does then you can separate verification into checking hashes and checking transitions and do those checks separately.

manta-trusted-setup/src/groth16/ppot/serialization.rs
GhostOfGauss: Who knows why I wrote it that way... anyway yeah it's the same thing as G2Affine so I've changed it to that.

manta-trusted-setup/src/groth16/ppot/serialization.rs
GhostOfGauss: Yeah that might be worth doing

manta-trusted-setup/src/groth16/ppot/serialization.rs
GhostOfGauss: Good catch. I probably originally wrote this function with that as an argument before realizing that the type `C` is what specifies the size of the subaccumulator. I've changed the documentation

manta-trusted-setup/src/groth16/ppot/serialization.rs
GhostOfGauss: Maybe so, and besides this test is really inelegantly written, I used a lot of copy-paste

manta-trusted-setup/src/groth16/ppot/kzg.rs
GhostOfGauss: You end up needing twice the number of G1 powers because there's a term of the form `h(X)t(X)` that can potentially have degree twice as large as the number of witnesses. Keeping in mind that `t(X)` is fixed by the circuit and `h(X)` changes from instance to instance, the prover key must contain commitments of the form `x^i t(x)`, which we compute [Here](https://github.com/Manta-Network/manta-rs/blob/3dbf4b5faec4d4a0c7f7c83d4dcd01528fbd8589/manta-trusted-setup/src/groth16/mpc.rs#L214-L217)

manta-trusted-setup/src/groth16/ppot/hashing.rs
SupremoUGH: I see, all clear. Do you think we should hash our small files anyway and store the values somewhere? This way we'd have some protection against tampering that doesn't involve going to the external files and checking their hashes.

manta-trusted-setup/Cargo.toml
```diff
@@ -42,18 +45,26 @@ std = ["manta-util/std"]
...
+ark-relations = { version = "0.3.0", default-features = false }
 ark-std = { version = "0.3.0", default-features = false }
 bincode = { version = "1.3.3", optional = true, default-features = false }
 blake2 = { version = "0.10.4", default-features = false }
-byteorder = { version = "1.4.3", default-features = false }
 derivative = { version = "2.2.0", default-features = false, features = ["use_core"] }
 manta-crypto = { path = "../manta-crypto", default-features = false, features = ["arkworks", "getrandom", "rand_chacha"] }
 manta-util = { path = "../manta-util", default-features = false }
+memmap = { version = "0.7.0", optional = true, default-features = false }
+rand_chacha = { version = "0.3.1", default-features = false }
```
bhgomes, line 59: We don't need the dependency anymore since `manta-crypto` handles it.

manta-trusted-setup/src/groth16/kzg.rs
```diff
@@ -253,6 +253,29 @@ where
                 .1,
         })
     }
+
+    /// Reinterpret a proof from a KZG ceremony as a proof
+    /// for a sub-ceremony.
+    #[inline]
+    pub fn cast_to_subceremony<D>(self) -> Proof<D>
+    where
+        D: Pairing<G1 = <C as Pairing>::G1, G2 = <C as Pairing>::G2>,
```
bhgomes, line 262: This should probably work:

```suggestion
        D: Pairing<G1 = C::G1, G2 = C::G2>,
```

manta-trusted-setup/src/groth16/ppot/hashing.rs
```diff
@@ -0,0 +1,226 @@
...
+mod tests {
+    use super::*;
+    use crate::groth16::{
+        kzg::{Accumulator, Proof},
+        ppot::{
+            kzg::PpotCeremony,
+            serialization::{read_kzg_proof, read_subaccumulator, Compressed, PpotSerializer},
+        },
+    };
+    use ark_std::println;
```
bhgomes, line 153: Just use `println!`.

manta-trusted-setup/src/util.rs
```diff
@@ -403,9 +411,12 @@ where
...
-    let mut seed = Vec::with_capacity(32);
-    for i in 0..8 {
-        seed.extend(BigEndian::read_u32(&digest[4 * i..4 * (i + 1)]).to_le_bytes());
+    let mut digest = digest.as_slice();
+    let mut seed = Vec::<u8>::with_capacity(32);
+    for _ in 0..8 {
+        let mut buffer = [0u8; 4];
+        let _ = digest.read(&mut buffer).unwrap();
+        seed.extend(buffer.iter().rev());
     }
```
bhgomes, line 420:
```suggestion
    let mut digest = digest.as_slice();
    let mut seed = Vec::with_capacity(32);
    for _ in 0..8 {
        let mut buffer = [0u8; 4];
        digest.read(&mut buffer).except("Reading into a slice never fails.");
        seed.extend(buffer.iter().rev());
    }
```

manta-trusted-setup/src/groth16/ppot/hashing.rs
```diff
@@ -0,0 +1,226 @@
...
+            .unwrap();
+        <PerpetualPowersOfTauCeremony<S, POWERS> as Serializer<G1Affine, G1>>::serialize_uncompressed(pair.1, &mut hasher)
+            .unwrap();
+        hash_to_group::<_, PpotDistribution, 64>(into_array_unchecked(hasher.0.finalize()))
+    }
+}
+
+/// A distribution to replicate random sampling as it was done
+/// during the Ppot ceremony, which used `rand v. 0.4`.
+#[derive(Copy, Clone, Default)]
```
bhgomes, line 61:
```suggestion
#[derive(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
```

manta-trusted-setup/src/groth16/ppot/hashing.rs
```diff
@@ -0,0 +1,226 @@
...
+    arkworks::{
+        ec::{short_weierstrass_jacobian::GroupAffine, ProjectiveCurve, SWModelParameters},
+        ff::{BigInteger256, Fp256, FpParameters, Zero},
+    },
+    rand::{RngCore, Sample},
+};
+use manta_util::into_array_unchecked;
+
+/// The G2 hasher used in the PPoT ceremony
+pub struct PpotHasher {
```
bhgomes, line 36:
```suggestion
#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
pub struct PpotHasher {
```

bhgomes (changes requested):

manta-trusted-setup/src/groth16/ppot/hashing.rs
GhostOfGauss: The CI doesn't allow that, it says the macro doesn't exist when it's checking with no std.  Using `ark_std` used to work and now for some reason it doesn't, so I've instead flagged the tests with `#[cfg(feature = "std")]`. Is that a good solution?

manta-trusted-setup/Cargo.toml
bhgomes: Sorry not `memmap` just `rand_chacha`.

manta-trusted-setup/src/groth16/ppot/hashing.rs
bhgomes: You should add `std` to the feature flags in the `dev-dependencies` for `manta-trusted-setup`.

SupremoUGH (approved):
Everything's resolve, approve.

bhgomes (approved):

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,962 @@
...
+    /// Element is a point on the G1 curve
+    fn is_g1_type(&self) -> bool {
+        !matches!(self, ElementType::TauG2 | ElementType::BetaG2)
+    }
+}
+
+/// Reads appropriate number of elements of `element_type` for an accumulator of given `Size` from PPoT challenge file.
+#[inline]
+pub fn read_g1_powers<S>(
+    readable_map: &Mmap,
```
bhgomes, line 570:
```suggestion
    reader: &[u8],
```

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,962 @@
...
+    S: Size,
+{
+    let size = element.num_powers::<S>();
+    let mut powers = Vec::new();
+    let mut start_position = calculate_mmap_position(0, element, compression);
+    let mut end_position = start_position + element.get_size(compression);
+    for _ in 0..size {
+        let mut reader = readable_map
+            .get(start_position..end_position)
+            .expect("cannot read point data from file");
```
bhgomes, line 584:
```suggestion
        let mut reader = reader[start_position..end_position];
```

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,962 @@
...
+                Compressed::Yes => {
+                    <PpotCeremony as Deserializer<G1Affine, G1>>::deserialize_compressed(
+                        &mut reader,
+                    )?
+                }
+            };
+            // point will be checked below
+            powers.push(point);
+        } else {
+            panic!("Expected G1 curve points")
```
bhgomes, line 602: Error?

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,962 @@
...
+        start_position = end_position;
+        end_position += element.get_size(compression);
+    }
+
+    // Do curve point checks in parallel
+    match compression {
+        Compressed::No => cfg_iter!(powers).for_each(|g| curve_point_checks(g).unwrap()),
+        Compressed::Yes => cfg_iter!(powers).for_each(|g| {
+            if !g.is_in_correct_subgroup_assuming_on_curve() {
+                panic!() // This should actually just return the NotInSubgroup error
```
bhgomes, line 613: Error?

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -0,0 +1,962 @@
...
+pub fn read_kzg_proof(readable_map: &Mmap) -> Result<Proof<PpotCeremony>, SerializationError> {
+    let position = 64
+        + (PpotCeremony::G1_POWERS + 2 * PpotCeremony::G2_POWERS) * 32
+        + (PpotCeremony::G2_POWERS + 1) * 64;
+    let reader = readable_map
+        .get(position..position + 6 * 64 + 3 * 128)
+        .expect("cannot read point data from file");
+
+    <Proof<_> as CanonicalDeserialize>::deserialize_uncompressed(reader)
+}
```
bhgomes, line 685:
```suggestion
#[inline]
pub fn read_kzg_proof(reader: &[u8]) -> Result<Proof<PpotCeremony>, SerializationError> {
    let position = 64
        + (PpotCeremony::G1_POWERS + 2 * PpotCeremony::G2_POWERS) * 32
        + (PpotCeremony::G2_POWERS + 1) * 64;
    Proof::<_>::deserialize_uncompressed(reader[position..position + 6 * 64 + 3 * 128])
}
```

bhgomes (changes requested):

manta-trusted-setup/Cargo.toml
```diff
@@ -42,18 +45,25 @@ std = ["manta-util/std"]
...
 ark-poly = { version = "0.3.0", default-features = false }
+ark-relations = { version = "0.3.0", default-features = false }
 ark-std = { version = "0.3.0", default-features = false }
 bincode = { version = "1.3.3", optional = true, default-features = false }
 blake2 = { version = "0.10.4", default-features = false }
-byteorder = { version = "1.4.3", default-features = false }
 derivative = { version = "2.2.0", default-features = false, features = ["use_core"] }
 manta-crypto = { path = "../manta-crypto", default-features = false, features = ["arkworks", "getrandom", "rand_chacha"] }
 manta-util = { path = "../manta-util", default-features = false }
+memmap = { version = "0.7.0", optional = true, default-features = false }
```
bhgomes, line 58: We can also move this to `dev-dependencies` now since we are moving to slices rather than memmaps.

manta-trusted-setup/src/groth16/ppot/serialization.rs
```diff
@@ -610,7 +609,8 @@ where
         Compressed::No => cfg_iter!(powers).for_each(|g| curve_point_checks(g).unwrap()),
         Compressed::Yes => cfg_iter!(powers).for_each(|g| {
             if !g.is_in_correct_subgroup_assuming_on_curve() {
```
SupremoUGH, line 611: question here: How do you check whether a point belongs to a subgroup without a discrete log?

SupremoUGH (approved):

manta-trusted-setup/src/groth16/ppot/serialization.rs
GhostOfGauss: Right, in principle you'd need a discrete log to prove that an arbitrary group element belongs to a particular cyclic subgroup. But if that subgroup has prime order `r` and is the only subgroup of order `r` then it is sufficient to check that `r * h = 0` to prove that `h` belongs to the given subgroup.  That's what this check does, it multiplies the point by `r` and checks that the result is 0.

GhostOfGauss: Now that I think more about it I suppose there's no need to assume `r` is prime in what I wrote above, it's enough for the group to have only one subgroup of order `r`. (Though in our case `r` is indeed prime)

bhgomes (approved):

PR #153, opened 2022-07-04
DO NOT MERGE: feat: upgrade to new circuits

bhgomes:
- [x] Separate `sender` and `receiver` modules from `transfer`
- [x] New UTXO Protocol
- [x] UTXO Protocol Abstraction
    - [x] Generalize `Transfer` over abstract UTXO Protocol
    - [x] Generalize `signer` over abstract `Transfer` Protocol
- [x] Upgrade concrete configuration
- [x] Upgrade testing and simulation framework
- [ ] change manta-parameters path back to `main`

---

Before we can merge this PR, please make sure that all the following items have been checked off:

- [x] Linked to an issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Added **one** line describing your change in [`CHANGELOG.md`](https://github.com/manta-network/manta-rs/blob/main/CHANGELOG.md) and added the appropriate `changelog` label to the PR.
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer.
- [x] Checked that changes and commits conform to the standards outlined in [`CONTRIBUTING.md`](https://github.com/manta-network/manta-rs/blob/main/CONTRIBUTING.md).

PR #197, opened 2022-08-09
Migrate Crypto Upgrades from #153

bhgomes:

## Changed Interfaces

- [x] `manta_crypto::accumulator`: Add `Types` base-trait for `accumulator` constructions. Now `MembershipProof` depends only on `M: Types` instead of `M: Model<COM>` so we can remove the compiler parameter and also add standard `derive` impls
- [x] `manta_crypto::algebra`: Rewrites the `Group` trait into the following set of traits `Group`, `Ring`, `ScalarMul`, `ScalarMulGroup`, and `HasGenerator`
- [x] `manta_crypto::constraint`: Replace `ProofSystemInput` with the `Input`/`HasInput` trait-pair (`Input` for implementation and `HasInput` for introspection)
- [x] `manta_crypto::eclair::num`: Expand `Zero` and `One` traits to match standard traits (including `zero` and `one` functions respectively)
- [x] `manta_crypto::permutation::duplex`: Add custom `Plaintext` and `Ciphertext` types to duplex encryption since `Vec<PlaintextBlock>` and `Vec<CiphertextBlock>` will not work (these are not fixed-size and cannot be allocated in the unknown case) we force the implementer to handle this instead of providing a half-solution. We also fix Poseidion encryption to handle this new correct allocation story

### Cryptography `manta_crypto`

- [x] `accumulator`: Remove unused `ItemHashAccumulatorModel`
- [x] `accumulator`: Remove unused `are_independent` from `Accumulator`
- [x] `arkworks::ff`: Add `try_into_XXX` integer conversion methods from `F: PrimeField` (with testing suite)
- [x] `constraint::measure`: Add `Measure` implementation for `()` the native compiler (currently no operations have a cost in this model, we can revisit this)
- [x] `constraint::measure`: Add `print_measurement` helper function to wrap any ECLAIR function call and print out it's measured size
- [x] `eclair`: Add `Type` introspection for the `Has` trait
- [x] `eclair::alloc`: Add `Constant` implementations for `bool`, `[T; N]`, `Vec<T>`, `Box<[T]>`, `Array<T, N>`, and `BoxArray<T, N>`
- [x] `eclair::alloc`: Add `Variable` implementations for `Box<T>`, `[T; N]`, `Array<T>`, `BoxArray<T>`
- [x] `eclair::alloc`: Generalize `Allocate` interface
- [x] `eclair::bool`: Implement `Assert` for `()`
- [x] `eclair::bool`: Add `ConditionalSelect` implementations for native integer types
- [x] `eclair::cmp`: Add `PartialEq` implementations for native integer types and for `Vec<T>`, `Box<[T]>`, `Array<T, N>`, `BoxArray<T, N>`
- [x] `eclair::num`: Add `Zero` and `One` implementations for integer types, `UnsignedInteger` and arkworks backend variables
- [x] `eclair::num`: Add `PartialEq`, `ConditionalSelect`, `ConditionalSwap` and `Variable` implementations for `UnsignedInteger`
- [x] `encryption`: Add `encryption::EmptyHeader` type for encryption schemes with no headers
- [x] `encryption`: Add `Encode` and `constraint::Input` implementations to `encryption::EncryptedMessage`
- [x] `encryption::hybrid`: Add `Variable` implementation to `hybrid::Randomness`, `hybrid::Ciphertext`
- [x] `encryption::hybrid`: Add `Encode` and `constraint::Input` implementations to `hybrid::Ciphertext`
- [x] `permutation::duplex`: Add `Variable`, `PartialEq`, `Encode`, and `constraint::Input` implementations to `duplex::Ciphertext`
- [x] `permutation::duplex`: Add `Constant` and `Sample` implementations to `duplex::Duplexer`
- [x] `rand`: Add `Sample` implementations for `PhantomData` and `()`
- [x] `signature`: Add more type aliases for signature schemes
- [x] `signature::convert`: Add signature message converters
- [x] `signature::schnorr`: Add `HasGenerator` implementation to `Schnorr` signature and simplify type bounds

### Accounting `manta_accounting`

- [x] `transfer`: Removed the `is_independent_from` methods for `Sender`/`Receiver` (unused)

### Utilities `manta_util`

- [x] `bytes`: Move `AsBytes` to `manta_util`
- [x] `codec`: Add more `Encode` implementations for native types
- [x] `convert`: Add `Field` and `Variant` introspection accessors
- [x] `cmp`: Add `Independence` relation which is a more general kind of non-equivalence relation. We will use it to distinguish between the kind of equivalence relation that checks if two nullifiers are strictly equal or their spend-receipt parts are equal (in the new protocol the `Nullifier = (commitment, outgoing_note)` and only the `commitment` is used to check the spend criterion so the independence relation checks if two `commitments` are unequal)
- [x] `iter`: Add `Iterable` extension traits for borrowed-iteration abstractions

## Miscelanous Updates

- [x] Expanded CI contexts
- [x] Expanded `NonNative` trait docs
- [x] Upgraded `clap`, `serde` and `anyhow`
- [x] Downgraded `bip32` (`v0.4.0` is incompatible with some node dependenices)

---

- [x] Ran `cargo hakari generate` to update the `workspace-hack` system


GhostOfGauss (approved):

manta-crypto/src/eclair/mod.rs
```diff
@@ -46,6 +46,22 @@ impl Native for () {
 /// This `trait` is explicitly not implemented for `()`, the default native compiler. This marker
 /// can be used to write explicitly different implementations for native and non-native compilers
 /// where otherwise a generic implementation would have to exist.
+///
+/// # Limitations
+///
+/// This is an emulation of an unimplemented feature of rust called ["negative trait bounds"]. As it
+/// currently stands, the compiler will make instances of [`NonNative`] outside of the ECLAIR crate
+/// unusable as you'll run into [error 0119] which has the following notice:
```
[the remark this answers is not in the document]
bhgomes: The main issue here is that `NonNative` (and its implementations for the native Rust types like `()`) is controlled by the `manta_crypto` crate. Adding an implementation for `NonNative` to `()` is not considered a breaking change by the Rust compiler since we're strictly increasing the number of interfaces that can interact with each other, not changing or removing existing ones. However, for a downstream crate that would use `NonNative` to give different implementations for `()` and all other `COM` arguments, this _is_ in fact a breaking change. In order to preserve the API breakage rules, Rust just does not allow the downstream crate to use `NonNative` as ECLAIR uses it.

manta-crypto/src/algebra.rs
```diff
@@ -56,15 +67,15 @@ pub trait Group<COM = ()> {
 )]
 #[derive(derivative::Derivative)]
 #[derivative(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
-pub struct DiffieHellman<G, COM = ()> {
+pub struct DiffieHellman<S, G> {
```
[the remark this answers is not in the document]
bhgomes: If we mention it here then we need to provide the `COM` argument which is just more verbose. It's better to avoid the `COM` argument on structures and keep them on `trait` definitions and `trait` implementations.

manta-crypto/src/signature/mod.rs
```diff
@@ -191,50 +209,68 @@ where
...
+        Debug(bound = "S: Debug, G: Debug"),
+        Eq(bound = "S: Eq, G: Eq"),
+        Hash(bound = "S: Hash, G: Hash"),
+        PartialEq(bound = "S: cmp::PartialEq, G: cmp::PartialEq")
     )]
-    pub struct Signature<G, COM = ()>
-    where
-        G: DiscreteLogarithmHardness + Group<COM>,
-    {
+    pub struct Signature<S, G> {
```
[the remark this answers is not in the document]
bhgomes: This is the same as the other point about adding the `COM` argument. We want to avoid adding it to this structure which would be needed. It's not `G: ScalarMulGroup<S>` that applies here but `G: ScalarMulGroup<S, COM>`.

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -298,7 +300,13 @@ where
         let _ = compiler;
```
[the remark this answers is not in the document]
bhgomes: Yeah, arkworks generates the constraints internally because every variable has a copy of the compiler. ECLAIR uses an explicit compiler for its interfaces so that's why it is an argument here. To use arkworks, we can just drop the borrow of the compiler and let the variables handle it (in fact there is nothing else we can do anyways).

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -380,18 +388,22 @@ where
...
 
+impl<C, CV> algebra::ScalarMul<ScalarVar<C, CV>, Compiler<C>> for GroupVar<C, CV>
+where
+    C: ProjectiveCurve,
+    CV: CurveVar<C, ConstraintField<C>>,
+{
     #[inline]
-    fn mul(&self, scalar: &Self::Scalar, compiler: &mut Compiler<C>) -> Self {
+    fn scalar_mul(&self, scalar: &ScalarVar<C, CV>, compiler: &mut Compiler<C>) -> Self {
         let _ = compiler;
```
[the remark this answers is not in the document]
bhgomes: Same as above. The constraints are added by arkworks internally. Here's an example: https://docs.rs/ark-r1cs-std/latest/src/ark_r1cs_std/fields/fp/mod.rs.html#175-192 (see line 185 where `enforce_constraint` is called).

manta-crypto/src/algebra.rs
```diff
@@ -27,25 +27,36 @@ use manta_util::codec::{Decode, DecodeError, Encode, Read, Write};
...
-    type Scalar: Scalar<COM>;
+/// Scalar Multiplication
+pub trait ScalarMul<S, COM = ()> {
+    /// Multiplies `self` by `scalar` in the group.
+    fn scalar_mul(&self, scalar: &S, compiler: &mut COM) -> Self;
+}
 
-    /// Adds `rhs` to `self` in the group.
-    fn add(&self, rhs: &Self, compiler: &mut COM) -> Self;
+/// Group with a Scalar Multiplication
```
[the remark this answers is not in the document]
bhgomes: Yeah, I'll update those thanks.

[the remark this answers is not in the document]
bhgomes: Actually it's weird that the CI did not catch this since some names have been removed.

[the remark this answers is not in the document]
bhgomes: Ok, it appears there was an old typo that prevented the CI to check these docs. Fixed now.

GhostOfGauss (dismissed):

BoyuanFeng (changes requested):
- TODO:
    - Add a parameter to `DiffieHellman` to support a generator with a different type than the public key group type
    - Add associated type to `HasGenerator` which returns a custom type

manta-crypto/src/accumulator.rs
```diff
@@ -147,7 +155,7 @@ where
 
     #[inline]
     fn prove(&self, item: &Self::Item) -> Option<MembershipProof<Self::Model>> {
-        (**self).prove(item)
+        (**self).prove(item).map(MembershipProof::into)
```
BoyuanFeng, line 158:
```suggestion
        (**self).prove(item)
```

manta-crypto/src/accumulator.rs
```diff
@@ -212,129 +220,36 @@ pub trait OptimizedAccumulator: Accumulator {
...
+    Debug(bound = "M::Witness: Debug, M::Output: Debug"),
+    Default(bound = "M::Witness: Default, M::Output: Default"),
+    Eq(bound = "M::Witness: Eq, M::Output: Eq"),
+    Hash(bound = "M::Witness: Hash, M::Output: Hash"),
+    PartialEq(bound = "M::Witness: PartialEq, M::Output: PartialEq")
+)]
+pub struct MembershipProof<M>
 where
-    M: Model<COM> + ?Sized,
+    M: Types + ?Sized,
```
BoyuanFeng, line 236: Why we put `M::Witness: Clone` in `Clone(bound = ...)` instead of having `witness: M::Witness + Clone`?

manta-crypto/src/arkworks/ff.rs
```diff
@@ -0,0 +1,118 @@
...
+        $(
+            #[doc = "Tries to convert a field element `x` into an integer of type `"]
+            #[doc = stringify!($type)]
+            #[doc = "`."]
+            #[inline]
+            pub fn $name<F>(x: F) -> Option<$type>
+            where
+                F: PrimeField,
+            {
+                if x < F::from(2u8).pow([$type::BITS as u64]) {
```
BoyuanFeng, line 36: What would happen if `$type::BITS as u64` is quite large such that `F::from(2u8).pow([$type::BITS as u64])` becomes a small integer after module?

We should have a comments to say `$type::BITS as u64` should be less than bits of `F`.

manta-crypto/src/constraint.rs
```diff
@@ -242,6 +274,29 @@ pub mod measure {
         }
     }
 
+    /// Prints the measurement of the call to `f` with the given `label`.
+    #[cfg(feature = "std")]
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "std")))]
+    #[inline]
+    pub fn print_measurement<COM, D, F, T>(label: D, f: F, compiler: &mut COM) -> T
```
BoyuanFeng, line 281:
```suggestion
    pub fn print_measurement<D, F, T, COM>(label: D, f: F, compiler: &mut COM) -> T
```

manta-crypto/src/eclair/alloc.rs
```diff
@@ -144,6 +224,90 @@ where
...
+        Self::new(compiler.allocate_unknown())
+    }
+
+    #[inline]
+    fn new_known(this: &Self::Type, compiler: &mut COM) -> Self {
+        Self::new(T::new_known(this, compiler))
+    }
+}
+
+impl<T, const N: usize, M, COM> Variable<M, COM> for [T; N]
```
BoyuanFeng, line 245:
```suggestion
impl<T, M, const N: usize, COM> Variable<M, COM> for [T; N]
```

manta-crypto/src/eclair/bool.rs
```diff
@@ -42,6 +42,13 @@ pub trait Assert: Has<bool> {
     }
 }
 
+impl Assert for () {
+    #[inline]
+    fn assert(&mut self, bit: &bool) {
+        assert!(bit)
+    }
+}
```
BoyuanFeng, line 50: When will we use `
let tmp = ();
tmp.assert(true);
`?

manta-crypto/src/eclair/cmp.rs
```diff
@@ -56,30 +60,155 @@ where
...
+        COM: Assert,
+    {
+        if self.len() != rhs.len() {
+            let not_equal = false.as_constant(compiler);
+            compiler.assert(&not_equal);
+        } else {
+            for (lhs, rhs) in self.iter().zip(rhs) {
+                compiler.assert_eq(lhs, rhs);
+            }
+        }
```
BoyuanFeng, line 118: Can we simply use `fn eq(&self, rhs: &Vec<Rhs>, compiler: &mut COM) -> Bool<COM>` and assert on its output?

manta-crypto/src/eclair/cmp.rs
```diff
@@ -56,30 +60,155 @@ where
...
+        COM: Assert,
+    {
+        if self.len() != rhs.len() {
+            let not_equal = false.as_constant(compiler);
+            compiler.assert(&not_equal);
+        } else {
+            for (lhs, rhs) in self.iter().zip(rhs.iter()) {
+                compiler.assert_eq(lhs, rhs);
+            }
+        }
```
BoyuanFeng, line 153: same above

manta-pay/Cargo.toml
```diff
@@ -105,10 +105,10 @@ ark-ed-on-bls12-381 = { version = "0.3.0", optional = true, default-features = f
 ark-groth16 = { version = "0.3.0", optional = true, default-features = false }
 ark-snark = { version = "0.3.0", optional = true, default-features = false }
 ark-std = { version = "0.3.0", optional = true, default-features = false }
-bip32 = { version = "0.4.0", optional = true, default-features = false, features = ["bip39", "secp256k1"] }
+bip32 = { version = "0.3.0", optional = true, default-features = false, features = ["bip39", "secp256k1"] }
```
BoyuanFeng, line 108: Why use an older version?

BoyuanFeng (changes requested):

manta-crypto/src/eclair/alloc.rs
bhgomes: Here `T` and `N` specify the type so they should go first.

manta-pay/Cargo.toml
bhgomes: See the explanation in the PR description.

manta-crypto/src/eclair/cmp.rs
bhgomes: This is more efficient from the point of view of # of constraints. This is the entire purpose of the `assert_equal` function. See the docstring.

manta-crypto/src/algebra.rs
```diff
@@ -55,8 +58,11 @@ pub trait HasGenerator<G, COM = ()>
 where
     G: Group<COM>,
 {
+    /// Generator Type
+    type Generator;
```
GhostOfGauss, line 62: I think it's strange for this trait to have the generic `G: Group` and not have `Generator` related to `G` in any way.  

Usually a generator is an element of the group that it generates, so that `G` and `Generator` would be the same type. The only exception I can think of at the moment is Lie groups, whose generators belong to Lie algebras and therefore do have a different type -- that's totally irrelevant here, but even in that case there's a clear relation between the type of the generator and the group it generates.

Anyway, this isn't a deal-breaker for me but I think it's worth asking if we want to say anything concrete about how `G` and the `Generator` type relate. Feel free to dismiss this as silly mathematical nit-picking :)

manta-crypto/src/eclair/alloc.rs
```diff
@@ -191,7 +355,12 @@ where
     }
 }
 
-impl<COM, T> Allocate<COM> for T where T: ?Sized {}
+impl<COM, T> Allocate<COM> for T
+where
+    COM: ?Sized,
+    T: ?Sized,
+{
+}
```
SupremoUGH, line 363: Naive question: doesn't this trait with implementation for any (possibly ?Sized) generic T allow us to implement both Constant and Variable for any T in an analogous way to what was done above, i.e., by using as_constant, as_known and as_unknown?

manta-crypto/src/eclair/cmp.rs
```diff
@@ -56,30 +60,155 @@ where
...
 
-/* TODO:
-impl<T, Rhs> PartialEq<Rhs> for T
+/// Implements [`PartialEq`] for the given `$type`.
+macro_rules! impl_partial_eq {
+    ($($type:tt),* $(,)?) => {
+        $(
+            impl<Rhs> PartialEq<Rhs> for $type
+            where
+                $type: cmp::PartialEq<Rhs>,
```
SupremoUGH, line 69: Are you implementing PartialEq for type, where type: PartialEq? Isn't this doing nothing?

manta-crypto/src/eclair/cmp.rs
```diff
@@ -56,30 +60,155 @@ where
...
+    fn assert_equal(&self, rhs: &BoxArray<Rhs, N>, compiler: &mut COM)
+    where
+        COM: Assert,
+    {
+        for (lhs, rhs) in self.iter().zip(rhs) {
+            compiler.assert_eq(lhs, rhs);
+        }
     }
 }
-*/
```
SupremoUGH, line 74: General comment on these implementations of PartialEq: it feels that they might be particular cases of a more general implementation, e.g., something like T : IntoIterator with T::Item : PartialEq.

manta-crypto/src/encryption/hybrid.rs
```diff
@@ -118,6 +122,28 @@ where
     }
 }
 
+impl<K, E, COM> Variable<Secret, COM> for Randomness<K, E>
+where
+    K: key::agreement::Types + Constant<COM>,
+    E: RandomnessType + Constant<COM>,
```
SupremoUGH, line 128: E: Constant + RandomnessType

manta-crypto/src/accumulator.rs
bhgomes: Not sure what this means. `witness: M::Witness` is a declaration of the field of a struct, you cannot put a trait here. The `Clone` bound is saying that if the `MembershipProof` type implements `Clone` it is because `M::Witness: Clone`.

manta-util/src/codec.rs
```diff
@@ -587,51 +587,65 @@ impl<T> Encode for PhantomData<T> {
...
+                    writer.write_ref(&self.to_le_bytes())?;
+                    Ok(())
+                }
+            }
+        )*
     }
 }
 
-impl Encode for u64 {
+encode_int!(u16, i16, u32, i32, u64, i64, u128, i128);
```
SupremoUGH, line 629: Why does u8 have a separate implementation? And where's i8?

SupremoUGH (changes requested):
Just a small remark: I didn't fully understand the `field_try_into` macro defined in `arkworks::ff`, so please check with someone else that everything's correct there.

manta-crypto/src/arkworks/ff.rs
bhgomes: Yeah, I can add a comment on the macro. It doesn't make sense to add it to the function.

manta-crypto/src/eclair/bool.rs
bhgomes: Remember that `()` is the native compiler so it needs to support as many functions as possible.

manta-crypto/src/algebra.rs
```diff
@@ -56,18 +73,18 @@ pub trait Group<COM = ()> {
 )]
 #[derive(derivative::Derivative)]
 #[derivative(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
-pub struct DiffieHellman<G, COM = ()> {
+pub struct DiffieHellman<S, G, GEN = G> {
```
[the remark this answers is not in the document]
bhgomes: Because `G: ScalarMul<S, COM>` is the correct bound but we don't have `COM` here. See the comment I wrote to Todd.

manta-crypto/src/permutation/duplex.rs
```diff
@@ -116,6 +135,83 @@ pub struct Ciphertext<T, C> {
...
 
+impl<T, C> Ciphertext<T, C> {
+    /// Builds a new [`Ciphertext`] from `tag` and `message`.
+    #[inline]
+    pub fn new(tag: T, message: C) -> Self {
+        Self { tag, message }
+    }
+}
+
+impl<T, C, COM> eclair::cmp::PartialEq<Self, COM> for Ciphertext<T, C>
```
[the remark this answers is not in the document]
bhgomes: Yes. `Self` will always equal the `Type` in `impl Trait for Type`.

manta-crypto/src/eclair/num.rs
```diff
@@ -180,6 +257,90 @@ where
...
+}
+
+impl<T, const BITS: usize, COM> ConditionalSelect<COM> for UnsignedInteger<T, BITS>
+where
+    COM: Has<bool>,
+    T: ConditionalSelect<COM>,
+{
+    #[inline]
+    fn select(bit: &Bool<COM>, true_value: &Self, false_value: &Self, compiler: &mut COM) -> Self {
+        Self::new_unchecked(T::select(bit, &true_value.0, &false_value.0, compiler))
```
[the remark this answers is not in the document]
bhgomes: This is a separate case as `zero` and `one` which are debatable whether they should be checked or not. This can be completely unchecked since the values `true_value` and `false_value` are already checked.

manta-crypto/src/eclair/num.rs
```diff
@@ -180,6 +257,90 @@ where
...
+
+impl<T, const BITS: usize, COM> ConditionalSwap<COM> for UnsignedInteger<T, BITS>
+where
+    COM: Has<bool>,
+    T: ConditionalSwap<COM>,
+{
+    #[inline]
+    fn swap(bit: &Bool<COM>, lhs: &Self, rhs: &Self, compiler: &mut COM) -> (Self, Self) {
+        let (lhs, rhs) = T::swap(bit, &lhs.0, &rhs.0, compiler);
+        (Self::new_unchecked(lhs), Self::new_unchecked(rhs))
```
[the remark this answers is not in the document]
bhgomes: This is a separate case as zero and one which are debatable whether they should be checked or not. This can be completely unchecked since the values true_value and false_value are already checked.

manta-util/src/codec.rs
bhgomes: It's separate because it's the canonical data type for raw data and the other integer types use it's implementation when defining their own so it cannot use the macro. I'll add an `i8`.

manta-crypto/src/eclair/cmp.rs
bhgomes: Notice the types here. `PartialEq` is the trait from ECLAIR and `cmp::PartialEq` is the trait from the Rust standard library. This macro lifts an implementation from Rust standard library to ECLAIR. If this were actually the same trait it would not compile since it would just be an infinite regress.

manta-crypto/src/eclair/alloc.rs
bhgomes: Because we give this implementation, this trait is in-fact _not_ implementable by anyone else (since we've given an implementation $\forall\text{ }\textsf{T}$). So for future implementers of `Constant` and `Variable` they will automatically inherit an `Allocate` implementation for their underlying `Constant::Type` or `Variable::Type` which comes with an `as_constant` and `as_known` methods. No one will be able to give a different implementation of these methods, they are just helper functions that are useful for reducing boilerplate.

manta-crypto/src/eclair/num.rs
```diff
@@ -180,6 +257,90 @@ where
...
 
+impl<T, const BITS: usize, COM> Zero<COM> for UnsignedInteger<T, BITS>
+where
+    T: Zero<COM>,
+{
+    type Verification = T::Verification;
+
+    #[inline]
+    fn zero(compiler: &mut COM) -> Self {
+        Self::new_unchecked(T::zero(compiler))
```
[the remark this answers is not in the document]
bhgomes: The idea here is that the zero value is less than `2^BITS` not that the zero value is representable in less than `BITS`-many bits. In the former case, it's clear that this is true.

manta-crypto/src/eclair/num.rs
```diff
@@ -180,6 +257,90 @@ where
...
+
+impl<T, const BITS: usize, COM> One<COM> for UnsignedInteger<T, BITS>
+where
+    T: One<COM>,
+{
+    type Verification = T::Verification;
+
+    #[inline]
+    fn one(compiler: &mut COM) -> Self {
+        Self::new_unchecked(T::one(compiler))
```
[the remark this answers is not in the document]
bhgomes: This is slightly more subtle. I'm not sure what to do here since its technically false that `1 < 2^0` so if `BITS` is set to zero then this is implementation is weird. One thing we could add is that `BITS > 0` whenever a compiler implements `AssertWithinBitRange`.

manta-crypto/src/eclair/num.rs
```diff
@@ -16,30 +16,107 @@
...
+        )*
+    }
+}
+
+define_zero_one!(u8, u16, u32, u64, u128, i8, i16, i32, i64, i128);
+
 /// Within-Bit-Range Assertion
 pub trait AssertWithinBitRange<T, const BITS: usize> {
-    /// Asserts that `value` is smaller than `2^BITS`
+    /// Asserts that `value` is smaller than `2^BITS`.
```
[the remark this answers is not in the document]
bhgomes: This is a compiler assertion, so you can think of this as living in the family of functions such as `assert` and `assert_eq`. These functions don't return anything but they modify the state of the compiler (in the case of R1CS they add constraints like `assert` => single bit constraint, `assert_eq` => equality constraint). For the native compiler, this reduces to a runtime assertion like the `assert!` macro.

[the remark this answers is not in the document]
bhgomes: Added an issue to improve this documentation in the future: https://github.com/Manta-Network/manta-rs/issues/204

manta-crypto/src/constraint.rs
```diff
@@ -84,12 +84,35 @@ pub trait ProofSystem {
 }
 
 /// Proof System Input
-pub trait ProofSystemInput<T>: ProofSystem
+pub trait Input<P>
+where
+    P: ProofSystem + ?Sized,
+{
+    /// Extends the `input` buffer with data from `self`.
```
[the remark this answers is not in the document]
bhgomes: Extend adds `self` to the public input. So if the public input is represented by `Vec<F>` and self is `(F, F)` then `extend` will be:

```rust
self.extend(&mut input) => { input.push(self.0); input.push(self.1); }
```

manta-crypto/src/permutation/duplex.rs
```diff
@@ -58,9 +71,15 @@ where
...
     type PlaintextBlock: Write<P, COM, Output = Self::CiphertextBlock>;
 
+    /// Plaintext Type
+    type Plaintext: FromIterator<Self::PlaintextBlock> + BorrowIterator<Self::PlaintextBlock>;
+
     /// Ciphertext Block Type
     type CiphertextBlock: Write<P, COM, Output = Self::PlaintextBlock>;
 
+    /// Ciphertext Type
+    type Ciphertext: FromIterator<Self::CiphertextBlock> + BorrowIterator<Self::CiphertextBlock>;
```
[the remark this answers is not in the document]
bhgomes: Yeah, this is the way to expose this `iter` method.

manta-crypto/src/algebra.rs
bhgomes: > I think it's strange for this trait to have the generic `G: Group` and not have `Generator` related to `G` in any way.  
> 
> Usually a generator is an element of the group that it generates, so that `G` and `Generator` would be the same type. The only exception I can think of at the moment is Lie groups, whose generators belong to Lie algebras and therefore do have a different type -- that's totally irrelevant here, but even in that case there's a clear relation between the type of the generator and the group it generates.
> 
> Anyway, this isn't a deal-breaker for me but I think it's worth asking if we want to say anything concrete about how `G` and the `Generator` type relate. Feel free to dismiss this as silly mathematical nit-picking :)

Yeah, you're still thinking too high-level, definitely didn't mean for this to have anything to do with Lie groups. `G` and `Generator` should actually be the same mathematical type but they need not be the same Rust types. For our case, we want `Generator` to be the pre-computed table and `G` to be the single group element, so although they really do mean the same mathematically, they don't mean the same in the code and in the performance characteristics (all the "side-effects" that are unobservable from a purely (classical) algebraic point-of-view). And it might be strange to think about, but it actually is _this `trait`_ that "relates" these two types together. The existence of this `trait` means that there is something that makes `Generator` relevant for operations with `G`. Actually, the part that's more strange is `G: Group<COM>`, `G` should probably get no type constraint at all. But we still want to make this about groups so that the word generator still has meaning here.

> Doesn't it make sense at this point to impose the extra constraint S: Ring? Also, maybe add a default implementation for any G and S = Z.

In general we want to reduce the number of constraints that are not _100%_ necessary for interface safety (and removing foot-guns). In this case `S` does not appear in this interface (to talk about groups we don't need to talk about scalar multiplication) and also $\mathbb{Z}$ is not a good default since it's not used very often (or even at all?) in cryptography.

manta-crypto/src/eclair/cmp.rs
bhgomes: We can't naively do that, but there is some compression we can do. Let's move to an issue: https://github.com/Manta-Network/manta-rs/issues/207.

manta-util/src/codec.rs
bhgomes: Actually, it doesn't have to be separate. I fixed it.

SupremoUGH (approved):
All my comments were addressed, approved.

manta-crypto/src/eclair/num.rs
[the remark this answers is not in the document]
bhgomes: I added a comment for this.

SupremoUGH (approved):

GhostOfGauss (approved):

manta-crypto/src/signature/mod.rs
```diff
@@ -191,50 +209,69 @@ where
...
+        type Group: ScalarMulGroup<Self::Scalar, COM, Output = Self::Group>
+            + DiscreteLogarithmHardness;
+
         /// Message Type
         type Message;
 
-        /// Hashes `message` along with `verifying_key` and `nonce_point` into a
-        /// [`Scalar`](Group::Scalar).
+        /// Hashes `message` along with `verifying_key` and `nonce_point` into a scalar of type
+        /// [`Scalar`](Self::Scalar.
```
BoyuanFeng, line 240:
```suggestion
        /// [`Scalar`](Self::Scalar).
```

BoyuanFeng (changes requested):
Only 1 minor typo.

BoyuanFeng (dismissed):

SupremoUGH (dismissed):

BoyuanFeng (approved):

SupremoUGH (approved):

PR #213, opened 2022-08-23
Add Ceremony Utilities

bhgomes:

Split out some of the utilities from the trusted setup PR: #177.

---

- [x] Ran `cargo hakari generate` to update the `workspace-hack` system


manta-crypto/src/dalek/ed25519.rs
```diff
@@ -0,0 +1,138 @@
...
+    let secret_key = generate_secret_key(rng);
+    Keypair {
+        public: (&secret_key).into(),
+        secret: secret_key,
+    }
+}
+
+/// Edwards Curve Signature Scheme for the `Curve25519` Elliptic Curve
+#[derive(derivative::Derivative)]
+#[derivative(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
BoyuanFeng, line 80: Why should we derive `Hash`, `Ord`, `PartialEq`, and `PartialOrd` for a signature function `ed25519`?

manta-crypto/src/signature/mod.rs
```diff
@@ -116,6 +121,62 @@ where
...
+}
+
+impl<T> SignedMessage<T>
+where
+    T: MessageType + SignatureType,
+{
+    /// Generates a new [`SignedMessage`] by signing `message` with `signing_key`.
+    #[inline]
+    pub fn new<COM>(
+        parameters: &T,
```
BoyuanFeng, line 158: imo, `parameters` is a signer to sign the `message` and generate a `signature`. Should we create a signer in this function instead of passing it as an argument?

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+impl<N> AsBytes for Message<N>
+where
+    N: AsBytes,
+{
+    #[inline]
+    fn as_bytes(&self) -> Vec<u8> {
+        let mut bytes = self.nonce.as_bytes();
+        bytes.append(&mut self.encoded_message.clone());
```
BoyuanFeng, line 69:
```suggestion
        bytes.extend_with_slice(&self.encoded_message);
```

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -0,0 +1,149 @@
...
+    type Error;
+}
+
+/// Signs the `message` with the `nonce` attached using the `signing_key`.
+#[inline]
+pub fn sign<T, S>(
+    signing_key: &S::SigningKey,
+    nonce: S::Nonce,
+    message: &T,
+) -> Result<S::Signature, bincode::Error>
```
BoyuanFeng, line 93: Use `serde_json` instead of `bincode::Error`?

manta-util/src/collections/btree_map.rs
```diff
@@ -0,0 +1,81 @@
...
+/// way to pop the last key-value pair out of a [`BTreeMap`] but requires cloning the key.
+#[inline]
+pub fn pop_last<K, V>(map: &mut BTreeMap<K, V>) -> Option<(K, V)>
+where
+    K: Clone + Ord,
+{
+    let key = map.keys().last()?.clone();
+    match map.remove(&key) {
+        Some(value) => Some((key, value)),
+        _ => unreachable!("{}", POP_LAST_ERROR_MESSAGE),
```
BoyuanFeng, line 42: Why panic instead of returning `None`?

manta-util/src/time/lock.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// resetting the modification time to [`Instant::now`].
+    #[inline]
+    pub fn mutate_timed<F, R>(&mut self, f: F) -> R
+    where
+        F: FnOnce(&mut T, Instant) -> R,
+    {
+        let result = f(&mut self.value, self.instant);
+        self.tap();
+        result
+    }
```
BoyuanFeng, line 105: Can we merge these two functions? Seems to not have significant difference.

manta-util/src/time/lock.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    /// Sets the internal value using `value` if the amount of time since the last modification to
+    /// `self` was longer than `timeout`.
+    #[inline]
+    pub fn set_with_if_expired<F>(&mut self, timeout: Duration, value: F) -> Option<T>
+    where
+        F: FnOnce() -> T,
+    {
+        self.mutate_if_expired(timeout, move |t| mem::replace(t, value()))
+    }
```
BoyuanFeng, line 136: What's the difference between these two functions?

workspace-hack/Cargo.toml
```diff
@@ -40,11 +40,13 @@ num-traits = { version = "0.2.15", features = ["i128", "libm", "std"] }
 pbkdf2 = { version = "0.9.0", default-features = false, features = ["hmac", "sha2"] }
 ppv-lite86 = { version = "0.2.16", default-features = false, features = ["simd", "std"] }
 rand = { version = "0.8.5", features = ["alloc", "getrandom", "libc", "rand_chacha", "std", "std_rng"] }
-rand_chacha = { version = "0.3.1", default-features = false, features = ["std"] }
-rand_core = { version = "0.6.3", default-features = false, features = ["alloc", "getrandom", "std"] }
+rand_chacha = { version = "0.3.1", default-features = false, features = ["serde", "serde1", "std"] }
+rand_core-d8f496e17d97b5cb = { package = "rand_core", version = "0.5.1", default-features = false, features = ["alloc", "getrandom", "std"] }
+rand_core-3b31131e45eafb45 = { package = "rand_core", version = "0.6.3", default-features = false, features = ["alloc", "getrandom", "std"] }
```
BoyuanFeng, line 45: Any idea why we have two versions of `rand_core` at the same time?

manta-util/src/time/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+//
+// manta-rs is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Temporal quantification.
```
BoyuanFeng, line 17:
```suggestion
//! Temporal quantification
```

manta-util/src/collections/btree_map.rs
```diff
@@ -0,0 +1,81 @@
...
+//
+// manta-rs is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! An ordered map based on a B-Tree.
```
BoyuanFeng, line 17:
```suggestion
//! An ordered map based on a B-Tree
```

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+//
+// manta-rs is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! A double-ended queue (deque) implemented with a growable ring buffer.
```
BoyuanFeng, line 17:
```suggestion
//! A double-ended queue (deque) implemented with a growable ring buffer
```

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+    all(feature = "serde-alloc", feature = "serde-array"),
+    derive(Deserialize, Serialize),
+    serde(
+        bound(deserialize = "T: Deserialize<'de>", serialize = "T: Serialize"),
+        crate = "crate::serde",
+        deny_unknown_fields
+    )
+)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct MultiVecDeque<T, const N: usize>(BoxArray<VecDeque<T>, N>);
```
BoyuanFeng, line 38: Why use `BoxArray` instead of `Array`?

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+impl<T, const N: usize> MultiVecDeque<T, N> {
+    /// Builds a new empty [`MultiVecDeque`].
+    #[inline]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns a shared reference to the [`VecDeque`] at the given `level`.
+    #[inline]
+    pub fn level(&self, level: usize) -> &VecDeque<T> {
```
BoyuanFeng, line 49: We may change the name `level` to `get_at_level`. The name `level` indicates that we want to get the level of something. But we are actually get something at a specific level.

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+        &self.0[level]
+    }
+
+    /// Returns a mutable reference to the [`VecDeque`] at the given `level`.
+    #[inline]
+    pub fn level_mut(&mut self, level: usize) -> &mut VecDeque<T> {
```
BoyuanFeng, line 55: Same above.

BoyuanFeng (changes requested):

manta-trusted-setup/src/groth16/ceremony/coordinator.rs
```diff
@@ -0,0 +1,113 @@
...
+
+/// Challenge Array Type
+pub type ChallengeArray<C, const N: usize> = BoxArray<Challenge<C>, N>;
+
+/// Participant Queue Type
+pub type Queue<C, const LEVEL_COUNT: usize> =
+    MultiVecDeque<<C as Ceremony>::Identifier, LEVEL_COUNT>;
+
+/// Ceremony Coordinator
+pub struct Coordinator<C, R, const CIRCUIT_COUNT: usize, const LEVEL_COUNT: usize>
```
BoyuanFeng, line 42: Consider derive Serialize/Deserialize

manta-crypto/src/dalek/ed25519.rs
bhgomes: It's a marker struct so it should derive all of the standard traits. We do this to make sure that these structures are compatible with interfaces that do not use correct derive semantics (like the standard Rust ones) and pose _extra_ unnecessary constraints on their types. Think something like the following:

```rust
#[derive(Clone, PartialEq)]
struct Object<T, S> {
    inner: T,
    __: PhantomData<S>,
}
```

This would also require that `S: Clone` or `S: PartialEq` whenever `Object<T, S>` needs to, which is the incorrect semantics. By setting marker structs to have all the standard traits we avoid this footgun for people who don't use the correct `derive` semantics because in this case if `S` were the `Ed25519` struct above, it would actually have implementations for all of those traits.

manta-crypto/src/signature/mod.rs
bhgomes: We should pass it by reference since the signature parameters can be shared amongst different objects and perhaps cannot be created. Think the same way as Poseidon parameters. We don't want to recreate them every time we use them, instead have one copy and pass around a pointer to it.

manta-trusted-setup/src/groth16/ceremony/coordinator.rs
```diff
@@ -0,0 +1,113 @@
...
+use manta_util::{collections::vec_deque::MultiVecDeque, time::lock::Timed, Array, BoxArray};
+
+/// Proof Array Type
+pub type ProofArray<C, const N: usize> = BoxArray<Proof<C>, N>;
+
+/// State Array Type
+pub type StateArray<C, const N: usize> = BoxArray<State<C>, N>;
+
+/// Challenge Array Type
+pub type ChallengeArray<C, const N: usize> = BoxArray<Challenge<C>, N>;
```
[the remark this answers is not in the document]
bhgomes: This is to avoid inefficiencies of using `Array` for large objects (stack overflow, slow iteration, etc). It's always safer to use boxed arrays, especially for these completely abstract types like `State` which could be very large.

manta-trusted-setup/src/groth16/ceremony/signature.rs
bhgomes: `extend_from_slice` but yes

manta-util/src/collections/btree_map.rs
bhgomes: Because we are guaranteed by the call on line `39` that this state is unreachable.

manta-util/src/collections/btree_map.rs
bhgomes: I wanted to keep the same docstring as the std library.

manta-util/src/collections/vec_deque.rs
bhgomes: Same as above.

manta-util/src/collections/vec_deque.rs
bhgomes: See other discussions above about `BoxArray` vs `Array`.

manta-util/src/time/lock.rs
bhgomes: We don't want to use `mutate_timed` if we don't care about the creation time, so we shouldn't burden the users of the API with that requirement on their functions.

manta-util/src/time/lock.rs
bhgomes: One of them executes the `value` function lazily only if the timer has expired, and the other one accepts the value. See for example the difference between the following:

- https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.resize
- https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.resize_with

workspace-hack/Cargo.toml
bhgomes: Just some internal part of the dependency tree, not sure why.

manta-util/src/time/mod.rs
bhgomes: Same as above about the std library docs.

manta-util/src/collections/vec_deque.rs
bhgomes: Yeah how about `at_level`.

manta-util/src/collections/vec_deque.rs
bhgomes: Yeah how about `at_level_mut`

manta-trusted-setup/src/groth16/ceremony/coordinator.rs
bhgomes: Yeah

manta-trusted-setup/src/groth16/ceremony/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+/// Participant
+pub trait Participant {
+    /// Participant Identifier Type
+    type Identifier;
+
+    /// Returns the [`Identifier`](Self::Identifier) for `self`.
+    fn id(&self) -> &Self::Identifier;
+
+    /// Returns the priority level for `self`.
+    fn level(&self) -> usize;
```
tsunrise, line 40: We might need to add a comment to explain how `level` related to priority. Based on current queue implementation (https://github.com/Manta-Network/manta-rs/pull/213/files#diff-08b386c96579d15a0ae6d6f894aabcc4d748f0cd59e834572906dc41c0a1052dR98-R100) lower lever indicates higher priority?

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn increment(&self) -> Self;
+
+    /// Checks if the current nonce is valid.
+    fn is_valid(&self) -> bool;
+}
+
+impl Nonce for u64 {
+    #[inline]
+    fn increment(&self) -> Self {
+        self.saturating_add(1)
```
tsunrise, line 35: Why `saturating_add`? Since `Self::MAX` nonce is invalid, we can do whatever we want here right (for example, `*self += 1`)?

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -0,0 +1,149 @@
...
+pub fn check_nonce<N>(current: &N, user_nonce: &N) -> bool
+where
+    N: Nonce,
+{
+    current.is_valid() && user_nonce.is_valid() && current == user_nonce
+}
+
+/// Message with Nonce
+#[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
+pub struct Message<N> {
```
tsunrise, line 55: we might need serialization here?

manta-trusted-setup/Cargo.toml
```diff
@@ -38,12 +41,12 @@ test = ["manta-crypto/test"]
 ark-groth16 = { version = "0.3.0", default-features = false }
 ark-poly = { version = "0.3.0", default-features = false }
 ark-std = { version = "0.3.0", default-features = false }
+bincode = { version = "1.3.3", optional = true, default-features = false }
```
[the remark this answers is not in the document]
tsunrise: We are using `bincode` for signing, which is a better way than `serde_json` imo

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -0,0 +1,149 @@
...
+where
+    T: Serialize,
+    S: SignatureScheme,
+{
+    S::default()
+        .verify(
+            verifying_key,
+            &Message {
+                nonce,
+                encoded_message: bincode::serialize(message)?,
```
tsunrise: @bhgomes Indeed I'm curious when to use `bincode` and when to use `ark_serialize`?

manta-util/src/time/lock.rs
```diff
@@ -0,0 +1,205 @@
...
+{
+    /// Builds a new [`Timed`] object from the default value of `T` and the current time returned by
+    /// [`Instant::now`].
+    #[inline]
+    fn default() -> Self {
+        Self::new(Default::default())
+    }
+}
+
+impl<T> Deref for Timed<T> {
```
tsunrise, line 191: Why not having `DerefMut`? Doing so, by calling `*object = something`, we set the value to be `something` and let the last modified time be `Instant::now()`. Is this because someone might get a mutable reference by calling `deref_mut` and set the value several seconds later but the last modified time is still the instant that `deref_mut` is called?

tsunrise (changes requested):

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -140,7 +141,7 @@ where
             verifying_key,
             &Message {
                 nonce,
-                encoded_message: bincode::serialize(message)?,
+                encoded_message: serde_json::to_vec(message)?,
```
tsunrise, line 144: @BoyuanFeng Why are we signing a json string instead of signing the bytes?? 🤔

manta-trusted-setup/src/groth16/ceremony/signature.rs
bhgomes: The serialization is handled inside the `sign` and `verify` functions this is more of an internal detail.

manta-trusted-setup/src/groth16/ceremony/signature.rs
bhgomes: Yeah, I agree with @tsunrise, lets's go back to `bincode`.

manta-util/src/time/lock.rs
bhgomes: Yeah, we want the time to be related to the time of the change, not the time of acquiring the mutable reference. There might be a way to do `DerefMut` but it's out of the scope for this PR. Not using `DerefMut` is definitely safer.

manta-trusted-setup/src/groth16/ceremony/signature.rs
bhgomes: We want to use `serde` (and concretely `bincode` and `serde_json`) for all serialization and `ark_serialize` is an internal serialization (implementation detail) that we have to use when serializing arkworks types. So we should avoid implementing `ark_serialize` ourselves (unless we are extending the arkworks interfaces), just consume objects that do so and adapt them to `serde`. It's not a clear cut distinction in all cases, but err on the side of using `serde`.

manta-trusted-setup/src/groth16/ceremony/mod.rs
bhgomes: Yeah good point, @BoyuanFeng can you add this comment?

manta-trusted-setup/src/groth16/ceremony/signature.rs
BoyuanFeng: I would like to suggest always implementing `serde` when we implement serialization by ourselves. We only use `ark_serialize` if it has been implemented in `arkworks`.

manta-trusted-setup/src/groth16/ceremony/signature.rs
BoyuanFeng: 1. I kind of feel there is no large difference between using `bincode` or `serde_json` since both `bincode` and `serde_json` use `serde`.
2. We can use one less crate if we only use `serde_json` instead of using both `serde_json` and `bincode`, since we use `serde_json` for all network messages,

bhgomes: Also, we shouldn't be using `serde_json` explicitly for network messages since `tide` and `reqwest` can do this automatically for us.

manta-trusted-setup/src/groth16/ceremony/signature.rs
bhgomes: Yeah exactly.

manta-trusted-setup/src/groth16/mpc.rs
```diff
@@ -53,6 +56,62 @@ pub type State<P> = ProvingKey<<P as Pairing>::Pairing>;
...
+    ///
+    /// This is also the size of the `b_g1_query` and the `b_g2_query`.
+    pub a_query: usize,
+
+    /// Size of the `h_query` in the [`ProvingKey`]
+    pub h_query: usize,
+
+    /// Size of the `l_query` in the [`ProvingKey`]
+    pub l_query: usize,
+}
```
SupremoUGH, line 82: Two comments:
first, the documentation is unclear, e.g., how do I know what `gamma_abc_g1` is without consulting the docs of `ark-groth16`? 
second, probably we shouldn't name the length of `gamma_abc_g1: usize` the same as the points themselves `gamma_abc_g1: Vec<E1::Affine>`.

manta-trusted-setup/src/util.rs
```diff
@@ -52,7 +51,7 @@ pub trait HasDistribution {
 /// [`Deserializer`] `trait`.
 ///
 /// [`CanonicalDeserialize`]: manta_crypto::arkworks::serialize::CanonicalDeserialize
-pub trait Serializer<T> {
+pub trait Serializer<T, M = ()> {
```
SupremoUGH, line 54: Why the new parameter?

manta-util/src/collections/btree_map.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use core::borrow::Borrow;
+
+#[doc(inline)]
+pub use alloc::collections::btree_map::*;
+
+/// Error Message for the [`pop_last`] Function
+const POP_LAST_ERROR_MESSAGE: &str =
+    "Querying the last element of the map must guarantee that `BTreeMap::remove` always succeeds.";
```
SupremoUGH, line 26: Why are you writing this message separately instead of just plugging it in `pop_last`?

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct MultiVecDeque<T, const N: usize>(BoxArray<VecDeque<T>, N>);
+
+impl<T, const N: usize> MultiVecDeque<T, N> {
+    /// Builds a new empty [`MultiVecDeque`].
+    #[inline]
+    pub fn new() -> Self {
+        Self::default()
+    }
```
SupremoUGH, line 45: Why doesn't new take any inputs?

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+        self.0.iter().map(VecDeque::len).sum()
+    }
+
+    /// Returns `true` if `self` has no elements.
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.0.iter().all(VecDeque::is_empty)
+    }
+
+    /// Returns a shared reference to the element at the given `index` in `self`.
```
SupremoUGH, line 71: Why use a single `index` instead of a multi-index (`level` + `index`)? Much more intuitive in my opinion.

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+                None => index -= level.len(),
+                item => return item,
+            }
+        }
+        None
+    }
+
+    /// Returns a mutable reference to the element at the given `index` in `self`.
+    #[inline]
+    pub fn get_mut(&mut self, mut index: usize) -> Option<&mut T> {
```
SupremoUGH, line 85: Same as above.

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+    {
+        self.is_front_with(item, PartialEq::eq)
+    }
+
+    /// Returns `true` if `item` is at the front of the deque using `eq` to compare elements of type
+    /// `T`.
+    #[inline]
+    pub fn is_front_with<F>(&self, item: &T, mut eq: F) -> bool
+    where
+        F: FnMut(&T, &T) -> bool,
```
SupremoUGH, line 122: Why `FnMut` and not `FnOnce`?

manta-util/src/collections/vec_deque.rs
```diff
@@ -0,0 +1,175 @@
...
+    {
+        self.position_with(level, item, PartialEq::eq)
+    }
+
+    /// Finds the position of `item` assuming it was inserted at the given `level` using `eq` to
+    /// compare elements of type `T`.
+    #[inline]
+    pub fn position_with<F>(&self, level: usize, item: &T, mut eq: F) -> Option<usize>
+    where
+        F: FnMut(&T, &T) -> bool,
```
SupremoUGH, line 145: Same as above.

SupremoUGH (changes requested):

manta-crypto/src/dalek/ed25519.rs
```diff
@@ -0,0 +1,138 @@
...
+            unreachable!("We are guaranteed the correct number of bytes from `SECRET_KEY_LENGTH`.")
+        }
+    }
+}
+
+/// Clones the `secret_key` by serializing and then deserializing.
+#[inline]
+pub fn clone_secret_key(secret_key: &SecretKey) -> SecretKey {
+    secret_key_from_bytes(secret_key.to_bytes())
+}
```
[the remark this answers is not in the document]
bhgomes: Because we don't own the `SecretKey` type.

manta-util/src/collections/vec_deque.rs
bhgomes: This function is there to get a specific `index` in the queue when you don't know it's level. You can get the element by level with `self.at_level(level).get(index)`.

manta-util/src/collections/vec_deque.rs
bhgomes: Because there is nothing to assign when creating a new container. It's like `Vec::new`.

manta-trusted-setup/src/groth16/mpc.rs
bhgomes: Yeah, I'd like to replace all of these with universal sizes like:

1. number of constraints
2. number of secret witness variables
3. number of public input variables

But this one for now at least matches the `VerifyingKey` and the entire infrastructure at this level relies on `ark-groth16` anyway. See this issue: https://github.com/Manta-Network/manta-rs/issues/214.

manta-trusted-setup/src/groth16/kzg.rs
```diff
@@ -35,6 +35,14 @@ use manta_util::{cfg_iter, cfg_iter_mut, from_variant, vec::VecExt};
 #[cfg(feature = "rayon")]
```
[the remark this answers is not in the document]
bhgomes: Because this one merges first and then we reduce the amount of code in that PR as well.

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -0,0 +1,150 @@
...
+
+//! Groth16 Trusted Setup Ceremony Signatures
+
+use alloc::vec::Vec;
+use manta_crypto::signature;
+use manta_util::{serde::Serialize, AsBytes};
+
+/// Nonce
+pub trait Nonce: PartialEq {
+    /// Increments the current nonce by one.
```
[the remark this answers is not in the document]
bhgomes: What would be the documentation change? We don't want `Nonce` to be automatically defined for existing types like those that implement `Add` since we need to take care that it satisfies the semantics of a nonce. It's safer to not rely on an existing interface.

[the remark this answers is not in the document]
bhgomes: In this case, we should actually consider instead actually an interface like

```rust
trait Nonce {
    fn increment(&self) -> Self;
    fn are_equal_and_valid(&self, rhs: &Self) -> bool;
}
```

so that there's no doubt. _cc: @BoyuanFeng_

manta-crypto/src/signature/mod.rs
BoyuanFeng: This `parameter` is the signer of type `T`, which is already generic.

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
+pub struct Message<N> {
+    /// Nonce
+    pub nonce: N,
+
+    /// Encoded Message
+    pub encoded_message: Vec<u8>,
+}
+
+impl<N> AsBytes for Message<N>
```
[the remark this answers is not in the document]
bhgomes: `AsBytes` has a single function that converts from `&self` to `Vec<u8>` whereas `serde::Serialize` is an entire framework for transforming types into any possible serialization system. For the `ed25519` signature we require at least `AsBytes` since they actually ask for `&[u8]` in our interface and that's too restrictive, we loosen the requirement to `AsBytes`.

manta-trusted-setup/src/util.rs
bhgomes: This is part of making it a markable trait that lets you give custom implementations even when `T` is the same type.

manta-trusted-setup/src/groth16/ceremony/registry.rs
```diff
@@ -0,0 +1,34 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Groth16 Trusted Setup Ceremony Registry
+
+/// Participant Registry
+pub trait Registry<I, P> {
```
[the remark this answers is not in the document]
BoyuanFeng: Registry is essentially a map from ID `I` to the participant `P`. So we want to keep both `I` and `P` generic.

manta-util/src/collections/vec_deque.rs
BoyuanFeng: `position` function requires FnMut(..), so we still need `FnMut` here.
`core::iter::traits::iterator::Iterator
pub fn position<P>(&mut self, predicate: P) -> Option<usize>
where
    Self: Sized,
    P: FnMut(Self::Item) -> bool,`

manta-trusted-setup/src/groth16/ceremony/signature.rs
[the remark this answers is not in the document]
BoyuanFeng: 1. We may not need `are_equal_and_valid` since we do not always check equality when we check validity.
2. `Add` trait enables addition with any number. We only need an increment of 1 for the nonce.

manta-util/src/collections/btree_map.rs
BoyuanFeng: This is a convention to put long and constant error message at the top of the code. If we have multiple such messages, it is easier to change without jumping to a specific line of code below and considering the code context.

manta-trusted-setup/src/groth16/ceremony/signature.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+/// Checks if the two nonces, `current` and `user_nonce`, are both valid and equal to each other.
+#[inline]
+pub fn check_nonce<N>(current: &N, user_nonce: &N) -> bool
+where
+    N: Nonce,
+{
+    current.is_valid() && user_nonce.is_valid() && current == user_nonce
```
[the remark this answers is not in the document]
BoyuanFeng: Suppose Nonce is `x mod 100` and `is_valid(x)` checks if x is smaller than 50, then `x1=5` and `x2=105` are equal but only `x1` is valid.

BoyuanFeng (approved):

SupremoUGH (approved):
All comments addressed, approve.

tsunrise (approved):

PR #215, opened 2022-08-23
Windowing algorithm for point multiplication

SupremoUGH:
New: 
`manta-crypto`:
- [x] `eclair::bool.rs`: Added `select_from_table` function with default implementation to `ConditionalSelect` trait.
- [x] `algebra.rs`: Added `add_assign`, `double_assign` and `repeated_double_assign` methods with default implementations to `Group` trait.
- [x] `algebra.rs`: Added `Window` struct. The method `scalar_mul` implements windowed multiplication.
- [x] `algebra.rs`: Added abstract correctness test `window_multiplication_correctness` for `scalar_mul`.

`manta-pay`:
- [x] `crypto::ecc::arkworks.rs`: Implemented the `PartialEq`, `ConditionalSelect` and `Zero` traits from eclair for `Group<C>` and `GroupVar`.
- [x] `crypto::ecc::arkworks.rs`: Added `windowed_mul_is_correct` test which tests a concrete implementation of `Window` for the native rust compiler.


manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -601,4 +649,20 @@ mod test {
...
+        let scalar = Scalar::<Bls12_381_Edwards>::gen(&mut OsRng);
+        let point = Group::<Bls12_381_Edwards>::sample((), &mut OsRng);
+        const WINDOW_SIZE: usize = 4;
+        window_multiplication_correctness(
+            WINDOW_SIZE,
+            &scalar,
+            point,
+            |scalar, _| scalar.0.into_repr().to_bits_be(),
+            compiler,
+        );
```
bhgomes, line 666: Also, can you add the circuit correctness check and the measurement check:
```suggestion
        window_multiplication_correctness(
            4,
            &Scalar::<Bls12_381_Edwards>::gen(&mut OsRng),
            Group::<Bls12_381_Edwards>::gen(&mut OsRng),
            |scalar, _| scalar.0.into_repr().to_bits_be(),
            &mut (),
        );
```

manta-pay/src/crypto/ecc/arkworks.rs
```diff
@@ -270,6 +274,50 @@ where
...
+impl<C> Zero for Group<C>
+where
+    C: ProjectiveCurve,
+{
+    type Verification = bool;
+
+    #[inline]
+    fn zero(compiler: &mut ()) -> Self {
+        let _ = compiler;
+        Group(<C::Affine as ark_std::Zero>::zero())
```
bhgomes, line 312: If this doesn't work, then use `ark_ff::Zero` instead, but don't use `ark_std` in general if you can (it's a cursed crate).

```suggestion
        Self(C::Affine::zero())
```

manta-crypto/src/eclair/bool.rs
```diff
@@ -96,6 +98,47 @@ where
...
+        B: IntoIterator<Item = &'s Bool<COM>>,
+        B::IntoIter: ExactSizeIterator,
+        T: IntoIterator<Item = &'s Self>,
+        T::IntoIter: ExactSizeIterator,
+    {
+        let mut table = table.into_iter();
+        let mut bits = bits.into_iter();
+        assert_eq!(
+            table.len(),
+            2usize.pow(bits.len() as u32),
```
bhgomes, line 119: Does the system complain about just having `2.pow(bits.len())` because otherwise just switch to that.

manta-crypto/src/algebra.rs
```diff
@@ -308,3 +430,37 @@ pub mod security {
...
+        scalar: &S,
+        point: G,
+        bit_conversion: F,
+        compiler: &mut COM,
+    ) where
+        G: ScalarMulGroup<S, COM, Output = G>
+            + Clone
+            + ConditionalSelect<COM>
+            + PartialEq<G, COM>
+            + Zero<COM>,
```
bhgomes, line 455: Sort these.

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+        &self.table
+    }
+
+    /// Returns the multiplication table, dropping `Window`.
+    #[inline]
+    pub fn into_inner(self) -> Vec<G> {
+        self.table
+    }
+
+    /// Doubles `result` `window_size` and then adds the elemnent from `table` corresponding to `chunk`.
```
bhgomes, line 195:
```suggestion
    /// Doubles `result`, `window_size`-many times, and then adds the element from `table` corresponding to `chunk`.
```

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
     }
 }
 
+/// Window Method for Point Multiplication
+pub struct Window<G> {
+    /// Multiplication Table
+    table: Vec<G>,
+}
```
bhgomes, line 148:
```suggestion
/// Group Element Table for Windowed Point Multiplication
#[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
pub struct Window<G> {
    /// Multiplication Table
    table: Vec<G>,
}
```

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+
+impl<G> Window<G> {
+    /// Creates a new `Window` from `table` without checking its correctness.
```
bhgomes, line 151:
```suggestion
    /// Creates a new [`Window`] from `table` without checking its correctness.
```

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+    #[inline]
+    pub fn new_unchecked(table: Vec<G>) -> Self {
+        Self { table }
+    }
+
+    /// Creates a new `Window`.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if `window_size` is less than `1`.
```
bhgomes, line 161:
```suggestion
    /// Creates a new [`Window`] table by repeatedly adding `point` to itself to
    /// support a table with windowed multiplication with `window_size`.
    ///
    /// # Panics
    ///
    /// This method panics if `window_size` is less than `1`.
```

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+    #[inline]
+    pub fn new<COM>(window_size: usize, point: G, compiler: &mut COM) -> Self
+    where
+        G: Group<COM> + Zero<COM>,
+    {
+        assert!(window_size > 0, "Window size must be at least 1.");
+        let table_length = 2usize.pow(window_size as u32);
```
bhgomes, line 168: Can this be `2.pow(window_size)`?

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+        for _ in 1..table_length {
+            table.push(table.last().unwrap().add(&point, compiler));
+        }
+        Self::new_unchecked(table)
+    }
+
+    /// Returns the window size.
+    #[inline]
+    pub fn window_size(&self) -> usize {
+        (self.table.len() as f32).log2() as usize
```
bhgomes, line 180:
```suggestion
        self.table.len().trailing_zeroes()
```

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+    }
+
+    /// Returns the multiplication table.
```
bhgomes, line 183:
```suggestion
    /// Returns a shared reference to the multiplication table.
```

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+    #[inline]
+    pub fn table(&self) -> &[G] {
+        &self.table
+    }
+
+    /// Returns the multiplication table, dropping `Window`.
```
bhgomes, line 189:
```suggestion
    /// Returns the multiplication table, dropping `self`.
```

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,100 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+        COM: Has<bool>,
+    {
+        let chunk = chunk.iter().copied().rev();
+        let selected_element = G::select_from_table(chunk, table, compiler);
+        result
+            .repeated_double_assign(window_size, compiler)
+            .add_assign(&selected_element, compiler);
+    }
+
+    /// Multiplies a point in G by `scalar` using `Window`.
```
bhgomes, line 214:
```suggestion
    /// Multiplies a point in `G` by `scalar` using `self` as the window table.
```

bhgomes (changes requested):

manta-crypto/src/eclair/bool.rs
SupremoUGH: It doesn't compile, requires a u32.

manta-crypto/src/algebra.rs
SupremoUGH: Done, although it doesn't derive Copy (I guess because Vec<T> doesn't implement copy even when T does).

manta-crypto/src/algebra.rs
SupremoUGH: It gives a compiler error in that case.

manta-crypto/src/algebra.rs
SupremoUGH: Done, but isn't this number terribly off if e.g. we have a (possibly unchecked) table length of 2^n + 1?

manta-pay/src/crypto/ecc/arkworks.rs
SupremoUGH: Done, circuit correctness and measurement checks to come.

manta-pay/Cargo.toml
```diff
@@ -102,6 +103,7 @@ websocket = [
 aes-gcm = { version = "0.9.4", default-features = false, features = ["aes", "alloc"] }
 ark-bls12-381 = { version = "0.3.0", optional = true, default-features = false, features = ["curve"] }
 ark-ed-on-bls12-381 = { version = "0.3.0", optional = true, default-features = false, features = ["r1cs"] }
+ark-ff = { version = "0.3.0", optional = true, default-features = false }
```
bhgomes, line 106: Use `manta_crypto::arkworks::ff` instead no need to import.

manta-crypto/src/eclair/bool.rs
```diff
@@ -96,6 +98,47 @@ where
 {
     /// Selects `true_value` when `bit == true` and `false_value` when `bit == false`.
     fn select(bit: &Bool<COM>, true_value: &Self, false_value: &Self, compiler: &mut COM) -> Self;
+
+    /// Selects an element from `table` by repeated iteration of `select` over `bits`.
+    /// The `bits` are ordered from most significant to least significant, forming unsigned integers in binary representation
+    /// which are understood as the `table` indices.
+    #[inline]
+    fn select_from_table<'s, B, T>(bits: B, table: T, compiler: &mut COM) -> Self
```
bhgomes, line 106: Check if https://docs.rs/ark-r1cs-std/latest/ark_r1cs_std/select/trait.CondSelectGadget.html#method.conditionally_select_power_of_two_vector is faster/slower than our implementation.

bhgomes (changes requested):

manta-crypto/src/eclair/bool.rs
SupremoUGH: Mathematical count is the same. Also, the primitive test

/// Runs our impl of table selection against arkwork's.
    #[test]
    fn constraints_per_algorithm() {
        let mut cs = Compiler::<Bls12_381_Edwards>::for_proofs();
        let scalar = Scalar::<Bls12_381_Edwards>::gen(&mut OsRng);
        let base = Group::<Bls12_381_Edwards>::sample((), &mut OsRng);
        let base_var =
            base.as_known::<Secret, GroupVar<Bls12_381_Edwards, AffineVar<_, _>>>(&mut cs);
        let scalar_var =
            scalar.as_known::<Secret, ScalarVar<Bls12_381_Edwards, AffineVar<_, _>>>(&mut cs);
        let window_size = 10usize;
        let window = Window::new(window_size, base_var, & mut cs).into_inner();
        let chunk = ToBitsGadget::to_bits_be(&scalar_var.0).unwrap();
        let bits = chunk.chunks_exact(window_size).next().unwrap();
        let window_2 = window.iter().map(|x| x.clone()).map(|x| x.0).collect::<Vec<_>>();
        let ctr1 = cs.constraint_count();
        let _my_algo = ConditionalSelect::select_from_table(bits, &window, & mut cs);
        let ctr2 = cs.constraint_count();
        let _their_algo = CondSelectGadget::conditionally_select_power_of_two_vector(bits, &window_2);
        let ctr3 = cs.constraint_count();
        assert!(cs.is_satisfied());
        println!("My algo: {:?}", ctr2 - ctr1);
        println!("Their algo: {:?}", ctr3 - ctr2);
    }
    
    returns the same number of constraints for both table look-ups. It seems it amounts to the same thing.

manta-crypto/src/eclair/bool.rs
tsunrise, line 119: Consider using `1 << bits.len()`?

manta-crypto/src/eclair/bool.rs
```diff
@@ -96,6 +98,47 @@ where
...
+            for bit in bits {
+                table = table
+                    .into_iter()
+                    .chunk_by()
+                    .map(|[x, y]| Self::select(bit, &y, &x, compiler))
+                    .collect();
+            }
+            table.take_first()
+        } else {
+            table
```
tsunrise, line 136: Are we assuming `bits` does not have trailing zeros? If so, add it to the comment.

tsunrise (changes requested):

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,111 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+        let bit_vector = bits.into_iter().collect::<Vec<_>>(); // TODO: Switch to chunking iterator.
+        let window_size = self.window_size();
+        let mut result = G::zero(compiler);
+        let mut iter_chunks = bit_vector.chunks_exact(window_size);
+        for chunk in &mut iter_chunks {
+            Self::scalar_mul_round(window_size, &self.table, chunk, &mut result, compiler);
+        }
+        let remainder = iter_chunks.remainder();
+        let last_window_size = remainder.len();
+        let subtable = &self.table[0..2usize.pow(last_window_size as u32)];
```
tsunrise, line 243: Consider using `1 << last_window_size`

manta-crypto/src/algebra.rs
```diff
@@ -308,3 +441,37 @@ pub mod security {
...
+            + Zero<COM>,
+        F: FnOnce(&S, &mut COM) -> B,
+        B: IntoIterator<Item = Bool<COM>>,
+        COM: Assert,
+    {
+        let product = point.scalar_mul(scalar, compiler);
+        let window = Window::new(window_size, point, compiler);
+        let bits = Vec::from_iter(bit_conversion(scalar, compiler)); // TODO: use iter instead of Vec.
+        let windowed_product = window.scalar_mul(&bits, compiler);
+        product.assert_equal(&windowed_product, compiler);
```
tsunrise, line 475: We might also need to check if all constraints are satisfied here?

tsunrise (changes requested):

manta-crypto/src/algebra.rs
SupremoUGH: Yes, you're right. However, I decided not to include it here because it'd be part of a separate PR I'm working on https://github.com/Manta-Network/manta-rs/pull/208, where I'll have to think about a nice framework for circuit constraint measurement and so on. 

Even if it didn't make it to the code for this very reason, I carried out a test in `manta-pay::ecc::arkworks.rs` imitating yours and cs.is_satisfied() turned true and the constraint measurement was the mathematically expected one (around 500 constraints in savings with a window of size 2).

tsunrise (approved):

bhgomes (changes requested):

bhgomes (dismissed):

bhgomes (approved):

manta-crypto/src/algebra.rs
```diff
@@ -113,6 +141,111 @@ impl<G, const N: usize> PrecomputedBaseTable<G, N> {
...
+        window_size: usize,
+        table: &[G],
+        chunk: &[&Bool<COM>],
+        result: &mut G,
+        compiler: &mut COM,
+    ) where
+        G: Clone + ConditionalSelect<COM> + Group<COM>,
+        COM: Has<bool>,
+    {
+        let chunk = chunk.iter().copied().rev();
```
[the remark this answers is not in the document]
SupremoUGH: On the contrary, `chunk` (the input of `scalar_mul_round`) goes from most to least significant. We take `rev()` here because `select_from_table` starts from the first bit and chooses the first/second element of each pair of the table, we want the least significant bit to do this first distinction. If you look at the test, you'll see that I use `to_bits_be`, so everything works in the big-endian representation. I'll add an implementation note for the `scalar_mul` method specifying that `bits` must be big-endian for it to be compatible with `select_from_table` and `Window::new`.

bhgomes (approved):

BoyuanFeng (approved):

PR #180, opened 2022-07-23
Add Arkworks Backend to `manta-crypto`

bhgomes:
- [x] Add Basic Crate Re-export and Canonical Dependencies
- [x] Add Sampling methods for fields and curves
- [x] Move some R1CS and algebra code from `manta-pay` to `manta-crypto`
- [x] Move `Pairing` abstraction from `manta-trusted-setup` to `manta-crypto`
- [x] Fix `cocoon` version dependency issue with a patch for now

---

Before we can merge this PR, please make sure that all the following items have been checked off:

- [x] Linked to an issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] Added **one** line describing your change in [`CHANGELOG.md`](https://github.com/manta-network/manta-rs/blob/main/CHANGELOG.md) and added the appropriate `changelog` label to the PR.
- [x] Re-reviewed `Files changed` in the GitHub PR explorer.
- [x] Checked that changes and commits conform to the standards outlined in [`CONTRIBUTING.md`](https://github.com/manta-network/manta-rs/blob/main/CONTRIBUTING.md).

Issue #40, opened 2022-03-21
Prepare `manta-crypto` for Export to Governing Organization


bhgomes:
After discussions with @stechu, we want to try and transition the `manta-crypto` tools to its own governing organization as a standalone public-good product (of the likes of `arkworks` and `zk-garage`). Before we do this, we should prepare the current code for export by finishing a few tasks:

- [ ] Backends
    - [ ] Design backend integration plan for `manta-crypto` and its (native and in-circuit) gadgets (with documentation)
    - [ ] Add `arkworks` backend
- [ ] Poseidon:
    - [ ] #22
    - [ ] Find the best way to move Poseidon into `manta-crypto` (requires _Backends_ first)
- [ ] ECC:
    - [ ] #24
    - [ ] #34
    - [ ] Move arkworks ECC to `manta-crypto` (requires _Backends_ first)

PR #248, opened 2022-09-08
Move `Fp` and `R1CS` to `manta-crypto`

GhostOfGauss:
Moves `Fp` and partially moves `R1CS` to `manta-crypto` as part of #40 

This copies `manta_pay::crypto::constraint::arkworks::codec` into `manta-crypto` without removing it from `manta-pay` because it is still referenced by too many other parts of `manta-pay`.  It will be removed in a future PR.

This removes `scale` entirely from `manta-pay`, its functionality is to be replaced in a way that @bhgomes can explain.

This finishes copying some `R1CS` impl's that weren't included in #180 but does not remove `R1CS` from `manta-pay` yet due to this being referenced by too much groth16 related code.  R1CS will be removed from manta-pay together with groth16 in a future PR.

---


manta-crypto/src/arkworks/codec.rs
```diff
@@ -0,0 +1,171 @@
...
+{
+    /// Builds a new [`ArkReader`] from `reader`.
+    #[inline]
+    pub fn new(reader: R) -> Self {
+        Self { state: Ok(reader) }
+    }
+
+    /// Updates the internal reader state by performing the `f` computation.
+    #[inline]
+    fn update<T, F>(&mut self, f: F) -> Option<T>
```
SupremoUGH, line 59: why do we want this fn to be private?

SupremoUGH (approved):
Everything seems fine, even though I don't fully understand (de)serialization, so please double check the details of that with Brandon.

manta-crypto/src/arkworks/codec.rs
```diff
@@ -0,0 +1,171 @@
...
+//
+// manta-rs is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with manta-rs.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Codec Utilities
```
bhgomes, line 17: The stuff in this file can go in `arkworks::serialize` instead, we don't need to split this out.

manta-crypto/src/arkworks/mod.rs
```diff
@@ -34,6 +34,7 @@ pub use ark_ed_on_bls12_381 as ed_on_bls12_381;
 pub use ark_ed_on_bn254 as ed_on_bn254;
 
 pub mod algebra;
+pub mod codec;
```
bhgomes, line 37: Like I mentioned before, we can use `serialize` here instead and do an extension of `ark-serialize` in that module.

manta-pay/src/config.rs
```diff
@@ -17,7 +17,7 @@
 //! Manta-Pay Configuration
 
 use crate::crypto::{
-    constraint::arkworks::{field_element_as_bytes, groth16, Boolean, Fp, FpVar, R1CS},
+    constraint::arkworks::{groth16, Boolean, R1CS}, // TODO: Move R1CS with Groth16
```
bhgomes, line 20: fix todo item or leave it out of here (use GH issue to track this)

manta-pay/src/crypto/constraint/arkworks/groth16.rs
```diff
@@ -19,7 +19,7 @@
 use crate::crypto::constraint::arkworks::{
     self,
     codec::{HasDeserialization, HasSerialization, SerializationError},
-    R1CS,
+    R1CS, // TODO: Move with Groth16
```
bhgomes, line 22: same as above TODO

manta-pay/src/crypto/constraint/arkworks/mod.rs
```diff
@@ -757,12 +451,13 @@ where
 /// Testing Suite
 #[cfg(test)]
 mod tests {
+    // TODO: move to constraint with R1CS
```
bhgomes, line 454: same as other TODO

bhgomes (changes requested):

manta-crypto/Cargo.toml
```diff
@@ -32,6 +32,7 @@ arkworks = [
     "ark-r1cs-std",
     "ark-relations",
     "ark-serialize",
+    "ark-std",
```
bhgomes, line 35: in this case, cause we need it, lets re-export it too

bhgomes (changes requested):

PR #249, opened 2022-09-09
adding bool_to_optin and const_fn_trait_bound for rust nightly

xgaozoyoe:
When compiling in nightly rust, it requires us enable bool_to_option feature. Also since we using trait function with non-constant bound "pub const fn scalar_bits<C>() -> usize" we need to add #![feature(const_fn_trait_bound)] for nightly rust.


bhgomes:
This feature is only needed for an old version of Rust, you can upgrade the stable compiler to get this feature. Our MSRV is just the latest version.

PR #264, opened 2022-10-07
client binary

### LINKED ISSUE #125: Trusted Setup Contribution CLI
TODO:

- [x] Add post-contribution message: "I have just made the {}th contribution to the Manta Network Trusted Setup Ceremony! The hash of my contribution is {}. #MantaTS"
- [x] Contribution Hash: $H(c||a||\pi) : \mathbb{B}^{32}$
- [x] ~~Add entropy messaging (with 5s window before entropy is sampled)~~

SupremoUGH:
New: 
- [x] Client binary
- [x] abstract client protocol in `client.rs`
- [x] concrete `Ceremony` implementation in `ppot` 

**Before approving this PR, please test the client binary exhaustively against the server on https://ceremony.manta.network, and report any unwanted behaviour.** 

## Bug

After some testing, I noticed a bug related to `Term::stdout`. When the terminal is on full-screen and you are queuing, it works fine. When the terminal is not on fullscreen, the method `clear_last_lines()` doesn't work and the client gets infinite `waiting in queue` messages. 
If you change the message to remove the last `\n`, then it works when not on fullscreen, but on FS it eats all lines above the queue message. 

How do we solve this?

Closes #125 

---


manta-trusted-setup/src/bin/groth16_phase2_client.rs
```diff
@@ -0,0 +1,109 @@
...
+                     |_|
+";
+
+/// Command
+#[derive(Debug, Subcommand)]
+pub enum Command {
+    /// Register for the Trusted Setup Ceremony
+    Register,
+
+    /// Runs the Trusted Setup Ceremony as a Contributor
```
BoyuanFeng, line 50:
```suggestion
    /// Contribute to the Trusted Setup Ceremony as a Contributor
```

manta-trusted-setup/src/groth16/ceremony/mod.rs
```diff
@@ -144,12 +175,56 @@ where
...
+                 Please make sure you have submitted your registration form and try again later.",
+            ),
+            Self::AlreadyContributed => {
+                write!(
+                    f,
+                    "You have already contributed to the ceremony. \
+                     Each participant is only allowed to contribute once.",
+                )
+            }
+            // TODO: Is this error reachable with our client?
```
BoyuanFeng, line 206: TODO

manta-trusted-setup/src/groth16/mod.rs
```diff
@@ -19,8 +19,11 @@
 pub mod kzg;
 pub mod mpc;
 
-#[cfg(all(feature = "bincode", feature = "serde"))]
-#[cfg_attr(doc_cfg, doc(cfg(all(feature = "bincode", feature = "serde"))))]
+#[cfg(all(feature = "bincode", feature = "serde", feature = "tokio"))]
```
BoyuanFeng, line 22: Should we have `tokio` here or for binary only?

manta-trusted-setup/src/groth16/mpc.rs
```diff
@@ -48,20 +56,62 @@ use {
...
-#[derivative(Clone(bound = ""))]
+#[derivative(Clone(bound = ""), Debug(bound = ""))]
 pub struct State<P>(
     #[cfg_attr(
         feature = "serde",
         serde(
-            serialize_with = "canonical_serialize::<ProvingKey<P::Pairing>, _>",
-            deserialize_with = "canonical_deserialize::<'de, _, ProvingKey<P::Pairing>>"
+            serialize_with = "canonical_serialize_uncompressed::<ProvingKey<P::Pairing>, _>",
+            deserialize_with = "canonical_deserialize_unchecked::<'de, _, ProvingKey<P::Pairing>>"
```
BoyuanFeng, line 65: Why use `...uncompressed` for serialize but `...unchecked` for deserialize?

BoyuanFeng (changes requested):

manta-trusted-setup/src/bin/groth16_phase2_client.rs
bhgomes: No, I think the original comment is better.

bhgomes:
> ## Bug
> 
> After some testing, I noticed a bug related to `Term::stdout`. When the terminal is on full-screen and you are queuing, it works fine. When the terminal is not on fullscreen, the method `clear_last_lines()` doesn't work and the client gets infinite `waiting in queue` messages. If you change the message to remove the last `\n`, then it works when not on fullscreen, but on FS it eats all lines above the queue message.
> 
> How do we solve this?

This seems resolved by the way.

manta-trusted-setup/src/groth16/mpc.rs
GhostOfGauss: The reason to use `deserialize_unchecked` is to prevent the client from performing on-curve/in-subgroup checks for the data it receives from the server -- this slows down the client too much, and we can consider the server to be trustworthy. We will add a manual check on the server side, since data from the client ought to be checked.  
  
Arkworks `deserialize_unchecked` assumes the serialization is uncompressed, which is why we need to use `serialize_uncompressed`

manta-trusted-setup/src/groth16/mod.rs
GhostOfGauss: I don't entirely understand why but removing `tokio` here makes the module `groth16::ceremony::client` to be unfindable by `groth16::ceremony::config::ppot`. The check `cargo clippy --no-default-features --features ppot-client` does not pass if we remove `tokio` here. Perhaps it could be moved somewhere else?

GhostOfGauss (approved):

BoyuanFeng (approved):

GhostOfGauss (approved):

BoyuanFeng (approved):

bhgomes (approved):

PR #267, opened 2022-10-10
TS Client Installation Script

GhostOfGauss:
Adds a convenient way for participants to download the client using `curl`. 

The script needs to detect the user's OS and download the appropriate version of the client.

---

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -18,36 +18,6 @@ fi
 
 set -u
 
-# # If RUSTUP_UPDATE_ROOT is unset or empty, default it.
-# RUSTUP_UPDATE_ROOT="${RUSTUP_UPDATE_ROOT:-https://static.rust-lang.org/rustup}"
-
-#XXX: If you change anything here, please make the same changes in setup_mode.rs
-# usage() {
-#     cat 1>&2 <<EOF
-# rustup-init 1.25.1 (48d233f65 2022-07-12)
-# The installer for rustup
-# 
-# USAGE:
-#     rustup-init [FLAGS] [OPTIONS]
-# 
-# FLAGS:
-#     -v, --verbose           Enable verbose output
-#     -q, --quiet             Disable progress output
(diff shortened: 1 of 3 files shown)


tools/install.sh
```diff
@@ -0,0 +1,668 @@
...
+    ensure downloader "$_url" "$_file" "$_arch"
+    ensure chmod u+x "$_file"
+
+    printf "DONE.\n"
+
+    case ":${PATH}:" in
+        *:"${_dir}":*)
+            ;;
+        *)
+            echo "\nexport PATH=\"${_dir}:\$PATH\";" >> "$HOME/.profile"
```
grenade, line 120: on my system this produces the following error:
```
$ source ~/.profile
bash: nexport: command not found...
```
which looks like a terminal quoting issue to me. a simple fix would be to add a single `echo` (no args) command on the line above which would produce the newline, then remove the `\n` from the start of this echo.

grenade (dismissed):

tools/install.sh
grenade: eg:
```bash
echo >> "$HOME/.profile"
echo "export PATH=\"${_dir}:\$PATH\";" >> "$HOME/.profile"
```

tools/install.sh
```diff
@@ -0,0 +1,668 @@
...
+# 
+# OPTIONS:
+#         --default-host <default-host>              Choose a default host triple
+#         --default-toolchain <default-toolchain>    Choose a default toolchain to install
+#         --default-toolchain none                   Do not install any toolchains
+#         --profile [minimal|default|complete]       Choose a profile
+#     -c, --component <components>...                Component name to also install
+#     -t, --target <targets>...                      Target name to also install
+# EOF
+# }
```
bhgomes, line 49: delete

bhgomes (changes requested):

bhgomes (approved):

PR #268, opened 2022-10-21
Trusted Setup Client v2

GhostOfGauss:
Updates the Trusted Setup Client to address two common problems:

- Invalid signatures in registration form: this problem is caused by participants incorrectly copying the relevant information (email, twitter, verifying key, signature) to our registration form. This problem is solved by creating customized links to the registration form that automatically fill in the above fields.
- Error message `Registry update is taking longer than expected...` for `contribute` mode was misleading. Participants commonly did not understand that they needed to wait until the ceremony begins to run the client in `contribute` mode. The error message is now more informative and leads them to [a landing page](https://ceremony.manta.network) for more information.

---


SupremoUGH (dismissed):
Should we keep the old messages as well? So people would get their public key + signature separately and then they have the link to the google form. Also this way we don't have to update the docs.

GhostOfGauss:
I prefer to keep the output of the client as simple as possible, so I think v2 should only display the link.

I can include a section in the docs that addresses v1 clients, but I plan to update the docs to assume v2.

manta-trusted-setup/src/groth16/ceremony/mod.rs
```diff
@@ -194,8 +194,8 @@ where
         match self {
             Self::NotRegistered => write!(
                 f,
-                "Registry update is taking longer than expected. \
-                 Please make sure you have submitted your registration form and try again later.",
+                "Please make sure you have submitted your registration form. \
+                 Check whether the ceremony has begun at <ceremony.manta.network>.",
```
bhgomes, line 198: This is consistent with the other address scheme above:

```suggestion
                 Check whether the ceremony has begun at https://ceremony.manta.network.",
```

manta-trusted-setup/src/groth16/ceremony/config/ppot.rs
```diff
@@ -333,18 +318,26 @@ pub fn register(twitter_account: String, email: String) {
...
+
+/// Generates link to registration form with Twitter, Email, Public Key,
+/// Signature fields pre-filled.
+pub fn register_link(
+    twitter: String,
+    email: String,
+    public_key: String,
+    signature: String,
+) -> String {
+    format!("https://mantanetwork.typeform.com/trustedsetup2#twitter={}&email={}&verifying_key={}&signature={} \n", twitter, email, public_key, signature)
```
bhgomes, line 340: I tried the form and it still asked me for my email. This is probably a bug right? Also, are we switching everyone to this new form (regardless of what version of the client they are using)? In that case I would just call it `trustedsetup` not `trustedsetup2`.

bhgomes (changes requested):
For the screenshot can use another one with the standard window size to see the wrapping (even if you wanted to use a wide one which I don't recommend, the one you posted has one character on the next line which is easy to miss).

manta-trusted-setup/src/groth16/ceremony/config/ppot.rs
GhostOfGauss: No, Jennifer wanted to keep the email question for reasons to do with integration with Mailchimp. This email address would be for our communications; the one provided by the link is what we would use for signature checking.

We're keeping both typeforms open and active in case some people keep v1 of the client and decide to resubmit registration, or share the v1 `.exe` with a friend, etc. So the old one is called `trustedsetup` and this new typeform is called `trustedsetup2`

bhgomes: Why would we have two different emails? It wasn't clear in the UI that they would be different.

manta-trusted-setup/src/groth16/ceremony/mod.rs
```diff
@@ -194,8 +194,8 @@ where
         match self {
             Self::NotRegistered => write!(
                 f,
-                "Registry update is taking longer than expected. \
-                 Please make sure you have submitted your registration form and try again later.",
+                "Please make sure you have submitted your registration form. \
+                 Check whether the ceremony has begun at https://ceremony.manta.network.",
```
SupremoUGH, line 198: I take people are getting this error because the box is running with our test server. Because otherwise redirecting them to ceremony.manta.network makes more sense when network errors happen.

SupremoUGH (approved):

bhgomes (approved):

PR #294, opened 2022-12-19
fix: distinguish between panic-errors and possible-fix-errors

bhgomes:

To ensure that panics don't occur in the runtime we have to distinguish between the errors that are non-recoverable (i.e. panics in Rust) and recoverable errors.

---

### DIFF
--- a/manta-accounting/src/transfer/mod.rs
+++ b/manta-accounting/src/transfer/mod.rs
@@ -1081,7 +1081,16 @@ where
 
     /// Error Type
     type Error: From<<Self as ReceiverLedger<Parameters<C>>>::Error>
-        + From<<Self as SenderLedger<Parameters<C>>>::Error>;
+        + From<<Self as SenderLedger<Parameters<C>>>::Error>
+        + Into<
+            TransferPostError<
+                C,
+                Self::AccountId,
+                <Self as SenderLedger<Parameters<C>>>::Error,
+                <Self as ReceiverLedger<Parameters<C>>>::Error,
+                <Self as TransferLedger<C>>::Error,
+            >,
+        >;
 
     /// Checks that the balances associated to the source accounts are sufficient to withdraw the
     /// amount given in `sources`.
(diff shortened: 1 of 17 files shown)


manta-pay/src/simulation/ledger/mod.rs
```diff
@@ -222,34 +222,57 @@ impl Ledger {
...
 
     #[inline]
-    fn is_unspent(&self, nullifier: Nullifier) -> Option<Self::ValidNullifier> {
+    fn is_unspent(
+        &self,
+        nullifier: Nullifier,
+    ) -> Result<Option<Self::ValidNullifier>, Self::Error> {
         if self.nullifiers.contains(&nullifier) {
-            None
+            Ok(None)
```
ghzlatarev, line 256: Shouldn't these Ok(None) be an error ?

SupremoUGH: It's a reasonable suggestion. I've chosen to keep the changes minimal, i.e., when we returned `None` before now we'll return `Ok(None)` and when we had an `expect` now we'll return an error type. Feel free to change the error types as you want if you think they'd be more appropriate.

ferrell-code (dismissed):

ghzlatarev (dismissed):
@ferrell-code not sure if the CI is expected to fail though

ferrell-code:
Yeah not sure either @ghzlatarev , but it seems to be an existing problem that was not introduced in this PR.

ferrell-code (dismissed):

ferrell-code (dismissed):

manta-accounting/src/transfer/mod.rs
```diff
@@ -1106,10 +1103,11 @@ where
         I: Iterator<Item = (Self::AccountId, C::AssetValue)>;
 
     /// Checks that the transfer proof stored in `posting_key` is valid.
+    #[allow(clippy::type_complexity)] // NOTE: Clippy is too harsh here.
     fn is_valid(
         &self,
         posting_key: TransferPostingKeyRef<C, Self>,
-    ) -> Option<(Self::ValidProof, Self::Event)>;
+    ) -> Result<Option<(Self::ValidProof, Self::Event)>, <Self as TransferLedger<C>>::Error>;
```
Garandor, line 1110: I kinda agree with @ghzlatarev that this would make more sense to just be a `Result<Proof,Err>` without the intermediate option, squelching clippy here is fixing the issue at the wrong end :)

ghzlatarev (dismissed):

ferrell-code (approved):

ferrell-code (approved):

PR #310, opened 2023-01-18
Identity verification

SupremoUGH:
Identity verification.
High level idea: we input an asset an a salt (which form a utxo), and we ask the user to generate a virtual (i.e. it won't go on the ledger) ToPublic transaction spending the utxo. To make sure that is the utxo that's being spent, we ask for a `UtxoMembershipProof` on top of the `TransferPost`.

- [x] Method `identity_proof` for signer.
- [x] Function `identity_verification`.
- [x] Testing 
---

### DIFF
--- a/manta-accounting/src/transfer/mod.rs
+++ b/manta-accounting/src/transfer/mod.rs
@@ -32,14 +32,15 @@
 use crate::{
     asset,
     transfer::{
+        canonical::TransferShape,
         receiver::{ReceiverLedger, ReceiverPostError},
         sender::{SenderLedger, SenderPostError},
-        utxo::{auth, Mint, NullifierIndependence, Spend, UtxoIndependence},
+        utxo::{auth, Mint, NullifierIndependence, Spend, UtxoIndependence, UtxoReconstruct},
     },
 };
 use core::{fmt::Debug, hash::Hash, iter::Sum, ops::AddAssign};
 use manta_crypto::{
-    accumulator::{self, ItemHashFunction, MembershipProof},
+    accumulator::{self, ItemHashFunction, MembershipProof, Model},
     constraint::{HasInput, Input, ProofSystem},
     eclair::{
         self,
@@ -2112,4 +2113,71 @@ where
             .iter()
             .for_each(|sink| C::ProofSystem::extend(input, sink.as_ref()));
     }
+}
+
+///
+pub enum IdentityVerificationError {
+    ///
+    InvalidShape,
+
+    ///
+    InvalidProof,
+
+    ///
+    InconsistentUtxoAccumulatorOutput,
+
+    ///
+    InvalidMembershipProof,
+}
+
+///
+#[allow(clippy::too_many_arguments)]
+#[inline]
+pub fn identity_verification<C>(
+    parameters: &Parameters<C>,
+    verifying_context: &VerifyingContext<C>,
+    utxo_accumulator_model: &UtxoAccumulatorModel<C>,
+    transferpost: TransferPost<C>,
+    utxo_membership_proof: UtxoMembershipProof<C>,
+    identifier: Identifier<C>,
+    asset: Asset<C>,
+    address: Address<C>,
+) -> Result<(), IdentityVerificationError>
+where
+    C: Configuration,
+    C::UtxoAccumulatorOutput: PartialEq,
+    UtxoAccumulatorModel<C>: Model<Verification = bool>,
+{
+    TransferShape::from_post(&transferpost).map_or(
+        Err(IdentityVerificationError::InvalidShape),
+        |shape| match shape {
+            TransferShape::ToPublic => Ok(()),
+            _ => Err(IdentityVerificationError::InvalidShape),
+        },
+    )?;
+    if !transferpost
+        .has_valid_proof(verifying_context)
+        .map_err(|_| IdentityVerificationError::InvalidProof)?
+    {
+        return Err(IdentityVerificationError::InvalidProof);
+    }
+    if !transferpost
+        .body
+        .sender_posts
+        .iter()
+        .any(|sender_post| &sender_post.utxo_accumulator_output == utxo_membership_proof.output())
+    {
+        return Err(IdentityVerificationError::InconsistentUtxoAccumulatorOutput);
+    }
+    let utxo = parameters.utxo_reconstruct(&asset, &identifier, &address);
+    if !utxo_membership_proof.verify(
+        utxo_accumulator_model,
+        &parameters
+            .utxo_accumulator_item_hash()
+            .item_hash(&utxo, &mut ()),
+        &mut (),
+    ) {
+        return Err(IdentityVerificationError::InvalidMembershipProof);
+    }
+    Ok(())
 }
--- a/manta-accounting/src/transfer/mod.rs
+++ b/manta-accounting/src/transfer/mod.rs
@@ -2195,7 +2195,7 @@ where
 pub fn identity_verification<A, C>(
     parameters: &Parameters<C>,
     verifying_context: &VerifyingContext<C>,
-    utxo_accumulator: &mut A,
+    utxo_accumulator_model: &UtxoAccumulatorModel<C>,
     identity_proof: IdentityProof<C>,
     virtual_asset: IdentifiedAsset<C>,
     address: Address<C>,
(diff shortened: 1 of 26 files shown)


manta-accounting/src/transfer/sender.rs
```diff
@@ -109,6 +109,27 @@ where
...
+    /// Samples a new [`PreSender`], returning also its [`Utxo`].
+    ///
+    /// [`Utxo`]: crate::transfer::utxo::UtxoType::Utxo
+    #[inline]
+    pub fn sample_with_utxo<R>(
+        parameters: &S,
+        authorization_context: &mut S::AuthorizationContext,
+        identifier: S::Identifier,
+        asset: S::Asset,
+        rng: &mut R,
```
ferrell-code, line 121: Doesn't randomness need to be generated outside of library

I think right now signer would generate the randomness not a verification server

Could be mistaken

manta-pay/src/test/signer.rs
```diff
@@ -0,0 +1,51 @@
...
+    let (proving_context, verifying_context, parameters, utxo_accumulator_model) =
+        load_parameters(directory.path()).expect("Failed to load parameters");
+    let mut signer = sample_signer(
+        &proving_context,
+        &parameters,
+        &utxo_accumulator_model,
+        &mut rng,
+    );
+    let identifier = Identifier::<Config>::new(false, rng.gen());
+    let virtual_asset = IdentifiedAsset::<Config>::new(identifier, rng.gen());
```
ferrell-code, line 37: More concretely, don't we want the randomness to be already generated here, no way to get deterministic output from this function for verification service

ferrell-code (dismissed):
I was a bit confused on flow but this looks like it should work to me.

manta-accounting/src/transfer/mod.rs
```diff
@@ -2110,3 +2115,133 @@ where
...
+    Debug(bound = "TransferPost<C>: Debug, UtxoMembershipProof<C>: Debug"),
+    Eq(bound = "TransferPost<C>: Eq, UtxoMembershipProof<C>: Eq"),
+    Hash(bound = "TransferPost<C>: Hash, UtxoMembershipProof<C>: Hash"),
+    PartialEq(bound = "TransferPost<C>: PartialEq, UtxoMembershipProof<C>: PartialEq")
+)]
+pub struct IdentityProof<C>
+where
+    C: Configuration,
+{
+    /// Transferpost
```
zqhxuyuan, line 2173: TransferPost

manta-pay/src/test/signer.rs
```diff
@@ -0,0 +1,51 @@
...
+    let address = signer.address();
+    let verification = identity_verification(
+        &parameters,
+        &verifying_context.to_public,
+        &utxo_accumulator_model,
+        identity_proof,
+        virtual_asset,
+        address,
+    );
+    assert!(verification.is_ok(), "Verification failed");
```
zqhxuyuan, line 50: add some failed case?

ferrell-code (approved):
nit

manta-accounting/src/wallet/signer.rs
```diff
@@ -1111,6 +1175,35 @@ where
...
+    /// Builds two virtual [`Sender`]s for `pre_sender`.
+    #[inline]
+    fn virtual_senders(
+        &mut self,
+        parameters: &Parameters<C>,
+        asset_id: &C::AssetId,
+        pre_sender: PreSender<C>,
+    ) -> Result<[Sender<C>; PrivateTransferShape::SENDERS], SignError<C>>
+    where
+        UtxoAccumulatorModel<C>: Clone,
```
bhgomes, line 1187: Instead of cloning, we should just have a function like

```rust
pub fn empty(model: &Self::Model) -> Self;
```

on the `Accumulator` trait that does the right thing.

manta-accounting/src/wallet/signer.rs
```diff
@@ -1111,6 +1175,35 @@ where
...
+        senders.push(sender);
+        let identifier = self.rng.gen();
+        senders.push(
+            self.build_pre_sender(
+                parameters,
+                identifier,
+                Asset::<C>::new(asset_id.clone(), Default::default()),
+            )
+            .upgrade_unchecked(Default::default()),
+        );
```
bhgomes, line 1203:
```suggestion
        senders.push(
            self.build_pre_sender(
                parameters,
                self.rng.gen(),
                Asset::<C>::new(asset_id.clone(), Default::default()),
            )
            .upgrade_unchecked(Default::default()),
        );
```

manta-accounting/src/wallet/signer.rs
```diff
@@ -1389,6 +1523,38 @@ where
...
+    #[inline]
+    pub fn identity_proof(
+        &mut self,
+        identified_asset: IdentifiedAsset<C>,
+    ) -> Option<IdentityProof<C>>
+    where
+        UtxoAccumulatorModel<C>: Clone,
+    {
+        let result = self.sign_virtual_to_public(identified_asset);
+        self.state.utxo_accumulator.rollback();
```
bhgomes, line 1537: We shouldn't need this anymore.

manta-accounting/src/transfer/utxo/mod.rs
```diff
@@ -100,6 +100,12 @@ pub trait IdentifierType {
 /// Identifier Type
 pub type Identifier<T> = <T as IdentifierType>::Identifier;
 
+/// Transparency
+pub trait IdentifierTransparency: IdentifierType {
+    /// Checks if `identifier` is transparent
+    fn is_transparent(identifier: &Self::Identifier) -> bool;
+}
```
bhgomes, line 107: Should remove the API

manta-accounting/src/transfer/mod.rs
```diff
@@ -2110,3 +2118,147 @@ where
...
+    ///
+    /// 1) `identity_proof.transfer_post` has a [`ToPublic`](crate::transfer::canonical::ToPublic)
+    /// shape.
+    ///
+    /// 2) `identity_proof.transfer_post` has a valid proof.
+    ///
+    /// 3) The [`UtxoAccumulatorOutput`] in `identity_proof.transfer_post` has been computed from
+    /// `virtual_asset` and `address`.
+    ///
+    /// 4) The virtual asset is correctly constructed.
```
bhgomes, line 2198: The body doesn't exactly follow this right? It should match.

bhgomes (changes requested):

manta-accounting/src/transfer/mod.rs
```diff
@@ -2110,3 +2118,147 @@ where
...
+                .utxo_accumulator_item_hash()
+                .item_hash(&utxo, &mut ()),
+        );
+        let utxo_accumulator_output = utxo_accumulator
+            .prove(
+                &parameters
+                    .utxo_accumulator_item_hash()
+                    .item_hash(&utxo, &mut ()),
+            )
+            .expect("Failed to generate UtxoMembershipProof")
```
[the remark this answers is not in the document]
SupremoUGH: Because at this level, we don't know that our `Accumulator` is a Merkle forest, so we can't use the `root()` method to get the `UtxoAccumulatorOutput`. The only way (at least that I could find) to obtain an output using only `accumulator.rs` methods is through the `prove` function.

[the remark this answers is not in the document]
bhgomes: Ok, then let's think of a way to extend the current APIs to get the root without having to build a proof that should be true for all accumulators. Perhaps something like

```rust
fn output_for(&self, item: &Self::Item) -> Option<Self::Output>;
```

manta-accounting/src/wallet/signer.rs
SupremoUGH: Doesn't let me because `build_pre_sender` is already taking a mutable reference of `self`.

ferrell-code (commented):
Can't comment much on correctness of crypto code, but `identity_verification` looks like it will work

manta-crypto/src/merkle_tree/tree.rs
```diff
@@ -1083,6 +1084,17 @@ where
         ))
     }
 
+    #[inline]
+    fn output_from(&self, item: &Self::Item) -> Option<Self::Output> {
+        let _ = item;
```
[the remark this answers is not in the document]
SupremoUGH: That's our convention in `manta-rs` for unused inputs so clippy doesn't complain.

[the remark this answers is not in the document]
ferrell-code: Oh ok, not a fan of  `output_from(&self, _item: &Self::Item)` ?

Not that it makes any difference to the compiler

ferrell-code (approved):

bhgomes (approved):

PR #314, opened 2023-01-31
Export Signer to WASM

SupremoUGH:
Goes together with https://github.com/Manta-Network/sdk/pull/78.

Changes: 
- Refactored most `SignerState` methods as free functions.
- Included `sign_with_transaction_data` compound method with a test. 
- Removed `std` from `wallet` feature because of compatibility issues with `sdk`. Changed the `AssetMap` from `HashMap` to `BTreeMap` as a consequence. 
- Several trait implementations (mostly `(de)Serialize` and `Hash`) for compatibility with `sdk`.
- Several type definitions.
- Removed `network` feature.

Misc: 
- Removed unused `utxo_utilities` module.
- Improved `TransactionData` verification method.

---


manta-accounting/src/wallet/signer/mod.rs
```diff
@@ -1533,50 +984,15 @@ where
     /// Returns the [`Address`] corresponding to `self`.
     #[inline]
     pub fn address(&mut self) -> Address<C> {
-        let account = self.state.accounts.get_default();
-        account.address(&self.parameters.parameters)
+        methods::address(self.parameters(), self.state.accounts())
```
zqhxuyuan, line 987: using `&self.parameters` and `&self.state.accounts` that align to other methods? or change other methods to `self.parameters()` and `self.state.accounts()`?

zqhxuyuan:
Mabye rename package name `methods` to `impl` or `util`?

manta-pay/Cargo.toml
```diff
@@ -112,7 +110,7 @@ websocket = [
 [dependencies]
 aes-gcm = { version = "0.9.4", default-features = false, features = ["aes", "alloc"] }
 bip0039 = { version = "0.10.1", optional = true, default-features = false }
-bip32 = { version = "0.3.0", optional = true, default-features = false, features = ["bip39", "secp256k1"] }
+bip32 = { version = "0.4.0", optional = true, default-features = false, features = ["bip39", "secp256k1"] }
```
bhgomes, line 113: Do you have any tests to confirm that this doesn't change the key values? I believe this caused issues before. Lets keep it at version 3 unless explicitly necessary for now.

manta-pay/src/signer/client/network.rs
```diff
@@ -53,6 +53,29 @@ impl Display for Network {
...
+/// Network Error
+#[cfg_attr(
+    feature = "serde",
+    derive(Deserialize, Serialize),
+    serde(crate = "manta_util::serde", deny_unknown_fields)
+)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub enum NetworkError {
+    /// Inexistent Wallet
+    InexistentWallet(Network),
```
bhgomes, line 76:
```suggestion
    /// Nonexistent Wallet
    NonexistentWallet(Network),
```

manta-pay/src/key.rs
```diff
@@ -288,6 +291,13 @@ impl PartialEq for Mnemonic {
     }
 }
 
+impl Hash for Mnemonic {
+    #[inline]
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.as_ref().hash(state)
+    }
+}
```
bhgomes, line 299: Why doesn't `derive(Hash)` work?

manta-crypto/src/arkworks/groth16.rs
```diff
@@ -211,6 +244,27 @@ where
...
+        self.proving_key.vk.gamma_abc_g1.hash(state);
+        self.proving_key.beta_g1.hash(state);
+        self.proving_key.delta_g1.hash(state);
+        self.proving_key.a_query.hash(state);
+        self.proving_key.b_g1_query.hash(state);
+        self.proving_key.b_g2_query.hash(state);
+        self.proving_key.h_query.hash(state);
+        self.proving_key.l_query.hash(state);
+    }
+}
```
bhgomes, line 266: Why doesn't `derive(Hash)` work?

manta-crypto/src/arkworks/groth16.rs
```diff
@@ -74,6 +80,18 @@ pub struct Proof<E>(
...
+where
+    E: PairingEngine,
+{
+    #[inline]
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.0.a.hash(state);
+        self.0.b.hash(state);
+        self.0.c.hash(state);
+    }
+}
```
bhgomes, line 93: Why doesn't `derive(Hash)` work?

bhgomes (changes requested):

manta-pay/src/key.rs
SupremoUGH: Because the `PartialEq` implementation is custom

manta-crypto/src/arkworks/groth16.rs
SupremoUGH: Because `ProvingKey` doesn't implement `Hash`, even though all its fields do.

manta-crypto/src/arkworks/groth16.rs
SupremoUGH: Same as above

manta-accounting/src/transfer/canonical.rs
```diff
@@ -738,4 +738,43 @@ where
...
+    pub fn verify(
+        &self,
+        parameters: &Parameters<C>,
+        address: &Address<C>,
+        transferpost: TransferPost<C>,
+    ) -> bool
+    where
+        Utxo<C>: PartialEq,
+    {
+        if !TransferShape::from_post(&transferpost)
```
ferrell-code, line 763: Does this actually verify the proofs? Doesn't it need the checks found here: https://github.com/Manta-Network/manta-rs/blob/6824ecbfaa4fbb7a7fa9b5fb558d8985954afb17/manta-accounting/src/transfer/mod.rs#L2199

Otherwise I could just make a phony `TransferPost<C>` and as long as UTXO is onchain it would seem valid

SupremoUGH: But the `TransferPost` is the challenge/public input provided by the verifier. It can be as phony as you want, then the prover will simply not be able to generate a valid proof. What needs to be verified is the `TransactionData`, and in that case, the verifier challenges the prover with a `TransferPost` and receives `TransactionData` from the verifier. You then verify it against the `TransferPost` you sent at the beginning. 

`identity_verification` works differently. The verifier provides the `IdentifiedAsset` and the prover generates a `TransferPost`. In this case, what needs to be verified is the `TransferPost` itself.

manta-pay/Cargo.toml
SupremoUGH: Apparently whatever was causing the incompatibility is no longer there.

bhgomes (dismissed):

bhgomes (approved):

PR #324, opened 2023-03-03
Security: Include the sink AccountId in the signed message for ToPublic transactions

SupremoUGH:
Due to the bug where an attacker could perform a replay attack on someone's valid ToPublic post by changing the public `AccountId`, we needed to sign this last field together with the `TransferPostBody`.

Main changes: 
-`auth::sign` and `auth::verify` now take `BodyWithAccountsRef` instead of `TransferPostBody`.
- transfer: 
1) `AccountId` is now an associated type of the `Configuration` trait, instead of the `Ledger` one.
2)  The `Transaction` enum in the `ToPublic` case now includes an `AccountId`.
3) `TransferPost` includes a `sink_accounts` field.
4) `IdentityVerification::verify` now verifies against a `public_account` as well.
- signer:
1) The function `identity_proof` also takes (a vector of) `C::AccountId` as input.

Misc:
- updated the `clap` dependency so it passes the lint after the latest rustup update.
- more on the `identity_proof` clumsy function: https://github.com/Manta-Network/manta-rs/issues/325

Related PRs: 
- [x] Update the pallet: https://github.com/Manta-Network/Manta/pull/1032
- [x] Update the sdk: https://github.com/Manta-Network/sdk/pull/102
- [x] Signer: https://github.com/Manta-Network/manta-signer/pull/332 
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 ### Security
+-[\#324](https://github.com/Manta-Network/manta-rs/pull/324) Include the sink AccountId in the signed message for ToPublic transactions.
 
 ## [0.5.11] - 2023-02-22
 ### Added
--- a/manta-accounting/src/transfer/canonical.rs
+++ b/manta-accounting/src/transfer/canonical.rs
@@ -338,8 +338,12 @@ impl TransferShape {
     derive(Deserialize, Serialize),
     serde(
         bound(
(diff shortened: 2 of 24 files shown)


Apokalip:
@SupremoUGH I don't think we should keep the public polkadot key in the State, was talking with Charlie. The key is needed only for signing so the easiest way in the perspective for current signer is to be provided from the front end in its request for signing as it can already see and process all the public accounts from polkadotjs. Keeping the public key in signer just complicates how we save history and state as right now state is constructed based on a KeySecret from a mnemonic(signer account). 

As for the signer-extension it would be easy to get and keep(cache) the keys as you have polkadotjs API directly or even the registed accounts in the extension itself.

stechu (approved):

PR #329, opened 2023-03-14
Initial signer sync

SupremoUGH:
Goes together with:
- https://github.com/Manta-Network/Manta/pull/1046
- https://github.com/Manta-Network/sdk/pull/108

Implements the Initial Synchronization method for the signer. For that, there's a new method that creates a partial Merkle tree from a vector of leaves and a path, extended to forked trees and forests. 
On the wallet side, we read the relevant data from the ledger and then use it to update the signer, without any Poseidon hash computation (as opposed to the usual sync method).

Design choice: 
- I chose to instantiate the MK from a path instead of a SinglePath MK. Using the second, we could check the roots of coincide, making it more secure, and less trusting on the pulled data. However, I think the data pulled from the Ledger is considered trusted, so no need for additional computations.

On the testing/simulation: Besides the new unit tests for the partial and forked MKs, I ran the simulation, kept the Ledger, and then re-ran it again with new actors which would do the initial sync w.r.t. the new ledger. It all went smoothly, as expected. The code was quite clumsy and it'd take a long time to make it look better (and add nice features like new actors joining the simulation at arbitrary times) so I removed that part, but it's good to know that it worked as expected. 

---


manta-accounting/src/wallet/mod.rs
```diff
@@ -265,6 +272,59 @@ where
...
+                .read(&checkpoint)
+                .await
+                .map_err(Error::LedgerConnectionError)?;
+            is_continue = should_continue;
+            request.extend_with_data(
+                &self
+                    .signer
+                    .transfer_parameters()
+                    .await
+                    .map_err(Error::SignerConnectionError)?,
```
bhgomes, line 318: you should grab these parameters once outside the loop

manta-accounting/src/wallet/signer/mod.rs
```diff
@@ -106,6 +122,171 @@ where
...
+)]
+pub struct InitialSyncData<C>
+where
+    C: transfer::Configuration + ?Sized,
+{
+    /// UTXO Data
+    pub utxo_data: Vec<Utxo<C>>,
+
+    /// Membership Proof Data
+    pub membership_proof_data: Vec<UtxoAccumulatorWitness<C>>,
```
bhgomes, line 169: Maybe just add a comment here that this is the proof of the latest leaves per tree.

manta-crypto/src/accumulator.rs
```diff
@@ -202,6 +203,23 @@ pub trait OptimizedAccumulator: Accumulator {
...
+    fn from_items_and_witnesses(
+        model: &Self::Model,
+        items: Vec<Vec<Self::Item>>,
+        witnesses: Vec<Self::Witness>,
+    ) -> Self;
+
+    /// Sorts `items`.
+    // TODO: move this to the model.
+    fn sort_items(items: Vec<Self::Item>) -> Vec<Vec<Self::Item>>;
+}
```
bhgomes, line 221: This needs a bit more clarification on what it does. And is it really sorting or more grouping?

manta-crypto/src/merkle_tree/forest.rs
```diff
@@ -525,6 +526,90 @@ where
...
+            #[inline]
+            fn sort_items(items: Vec<Self::Item>) -> Vec<Vec<Self::Item>> {
+                let mut result = Vec::<Vec<Self::Item>>::default();
+                result.resize_with(N, Default::default);
+
+                for item in items {
+                    let tree_index = C::tree_index(&item).into();
+                    result[tree_index].push(item);
+                }
+
```
bhgomes, line 594:
```suggestion
                for item in items {
                    let tree_index = C::tree_index(&item).into();
                    result[tree_index].push(item);
                }
```

bhgomes (changes requested):

Apokalip (approved):
lgmt

PR #330, opened 2023-03-24
Batch insert

SupremoUGH:
## Description: 

Adds batch insertion methods to the Merkle Tree implementations, in particular to `Full`, `Partial`, `Fork` and `Forest`. Implements batch insertion as the default for the sync protocol.

---


Apokalip (dismissed):

manta-crypto/src/merkle_tree/full.rs
```diff
@@ -209,6 +227,49 @@ where
         self.push_leaf_digest(parameters, Node(len), leaf_digest()?);
         Some(true)
     }
+
+    #[inline]
+    fn batch_maybe_push_digest<F>(
```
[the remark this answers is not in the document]
SupremoUGH: This is an impl of a trait, we don't add comments in this case.

manta-crypto/src/accumulator.rs
```diff
@@ -203,6 +203,57 @@ pub trait OptimizedAccumulator: Accumulator {
...
+    /// must return the same result as the default implementation.
+    ///
+    /// [`insert_nonprovable`]: OptimizedAccumulator::insert_nonprovable
+    #[inline]
+    fn batch_insert_nonprovable<'a, I>(&mut self, items: I) -> bool
+    where
+        Self::Item: 'a,
+        I: IntoIterator<Item = &'a Self::Item>,
+    {
+        for item in items.into_iter() {
```
bhgomes, line 248:
```suggestion
        for item in items {
```

manta-crypto/src/merkle_tree/test/batch_insertion.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// Runs [`test_batch_insertion`] on a [`ForkedTree`].
+#[test]
+fn test_batch_insertion_fork() {
+    test_batch_insertion(|parameters| ForkedTree::new(Partial::new(parameters), parameters))
+}
+
+/// Tests batch insertion on a Merkle forest.
+#[test]
+fn test_batch_insertion_forest() {
```
bhgomes, line 97: Why can't you use the same thing as the tree code? If the types can't match up at least use a macro to generate both of these functions to make sure they stay consistent.

manta-crypto/src/merkle_tree/test/batch_insertion.rs
```diff
@@ -0,0 +1,115 @@
...
+    let mut cloned_tree = tree.clone();
+    let number_of_insertions = rng.gen_range(1..(1 << (HEIGHT - 1)));
+    let mut insertions = Vec::with_capacity(number_of_insertions);
+    for _ in 0..number_of_insertions {
+        insertions.push(rng.gen());
+    }
+    for leaf in &insertions {
+        tree.push(&parameters, leaf);
+    }
+    cloned_tree.batch_push(&parameters, &insertions);
```
bhgomes, line 70: Is there a need to test `batch_insert` vs `batch_insert_nonprovable` at the accumulator level? Maybe you can write an abstract test in the accumulator module, that should also deduplicate code here.

manta-crypto/src/merkle_tree/forest.rs
```diff
@@ -593,6 +593,60 @@ macro_rules! impl_from_items_and_witnesses {
...
+            C::Index: FixedIndex<N>,
+            Parameters<C>: Clone,
+            Leaf<C>: Clone,
+            LeafDigest<C>: Clone + Default + PartialEq,
+            InnerDigest<C>: Clone + Default + PartialEq,
+        {
+            /// # Note
+            ///
+            /// This implementation does not yield the same result as individually inserting
+            /// each item in `items` with [`insert`](Accumulator::insert) till failure.
```
bhgomes, line 609: Why wouldn't it yield the same result? Each tree is independent, so it shouldn't matter what order you compute these in.

manta-crypto/src/merkle_tree/node.rs
```diff
@@ -159,7 +160,7 @@ impl Default for Parity {
     derive(Deserialize, Serialize),
     serde(crate = "manta_util::serde", deny_unknown_fields)
 )]
-#[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
bhgomes, line 163: Why is the canonical ordering the right one? Maybe explain in the doc for this type.

manta-crypto/src/merkle_tree/node.rs
```diff
@@ -433,3 +440,117 @@ impl Iterator for NodeParents {
 }
 
 impl FusedIterator for NodeParents {}
+
+/// Dual Parity
```
bhgomes, line 444: Just add a use case here.

manta-crypto/src/merkle_tree/node.rs
```diff
@@ -433,3 +440,117 @@ impl Iterator for NodeParents {
...
+    serde(crate = "manta_util::serde", deny_unknown_fields)
+)]
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct DualParity(pub (Parity, Parity));
+
+impl DualParity {
+    /// Returns the starting [`Parity`] of `self`.
+    #[inline]
+    pub const fn starting_parity(&self) -> Parity {
+        self.0 .0
```
bhgomes, line 457:
```suggestion
        (self.0).0
```

bhgomes (changes requested):
Initial Review: I still have 4 more files to read.

manta-crypto/src/merkle_tree/node.rs
SupremoUGH: I used `Ord` for an earlier design, it's not needed for the final one, so I just removed it.

manta-crypto/src/merkle_tree/forest.rs
SupremoUGH: It doesn't yield the same result as the default impl of `BatchInsertion` for `Forest`, which would stop after the first failed insertion (even though there may be trees with empty slots left). This custom implementation only halts the insertions tree-wise, which makes more sense. I'm only adding the comment because the implementation doesn't 100% adhere to the comment in the `BatchInsertion` trait, but it is strictly better.

manta-crypto/src/merkle_tree/test/batch_insertion.rs
SupremoUGH: No need, those functions are the same as `batch_push`. The only reason I didn't is `ForkedTree` doesn't implement `Accumulator` so it wouldn't fit in the generic function above.

Apokalip (approved):

bhgomes (approved):

PR #331, opened 2023-03-31
Merkle tree pruning

SupremoUGH:
Implements Merkle tree pruning and nonprovable insertion.
Goes together with: https://github.com/Manta-Network/sdk/pull/118

Misc: 
- fixes a bug with the initial sync where the loop would be stuck because the checkpoint wasn't getting updated.
- improves the signer key logic

---

### DIFF
--- a/manta-accounting/src/wallet/mod.rs
+++ b/manta-accounting/src/wallet/mod.rs
@@ -272,6 +272,36 @@ where
         Ok(())
     }
 
+    /// Pulls data from the ledger, synchronizing the wallet and balance state. This method loops
+    /// continuously calling [`sbt_sync_partial`](Self::sbt_sync_partial) until all the ledger data has
+    /// arrived at and has been synchronized with the wallet.
+    ///
+    /// # Failure Conditions
+    ///
+    /// This method returns an element of type [`Error`] on failure, which can result from any
+    /// number of synchronization issues between the wallet, the ledger, and the signer. See the
+    /// [`InconsistencyError`] type for more information on the kinds of errors that can occur and
+    /// how to resolve them.
+    ///
+    /// # Note
+    ///
+    /// In general, this method does not update the [`Utxo`] accumulator, thus making the new assets
+    /// effectively non-spendable. Therefore, this method should only be used when the pallet does not
+    /// allow [`PrivateTransfer`]s or [`ToPublic`] transactions, for example in the case of
+    /// Soul-Bound Tokens (SBTs).
+    ///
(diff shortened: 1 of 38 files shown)


manta-accounting/src/wallet/mod.rs
```diff
@@ -339,7 +339,7 @@ where
             let ReadResponse {
                 should_continue,
                 data,
-            } = self.read_from_ledger().await?;
+            } = self.read_from_ledger(&checkpoint).await?;
```
zqhxuyuan, line 342: the checkpoint here at first will be Default(0), so every time the wallet trying to call `initial_sync`, it will sync from beginning? So I suppose this initial_sync is only called once when user first time install the wallet. What if usre disable the wallet and enable it again, do we call initial_sync again? @DanielZhangReal

DanielZhangReal: It will only be called once, and the result will be saved to indexedDB after success. 
Disabling and enabling will not affect indexedDB. 
Also, this logic only exists when the user creates a wallet, not the import logic

zqhxuyuan (dismissed):

manta-pay/src/simulation/mod.rs
```diff
@@ -72,7 +72,6 @@ where
         utxo_accumulator_model,
     );
     signer.load_accounts(AccountTable::new(KeySecret::sample(rng)));
-    signer.update_authorization_context();
```
bhgomes, line 75: Why was this removed?

manta-crypto/src/merkle_tree/partial.rs
```diff
@@ -365,12 +377,64 @@ where
         self.push_leaf_digest(parameters, Node(len), leaf_digest()?);
         Some(true)
     }
+
+    /// Removes the paths corresponding to the nonprovable leaves in `self`.
+    #[inline]
+    pub fn prune(&mut self) {
+        if let Some(current_index) = self.leaf_map.current_index() {
+            let marked_indices = (0..current_index)
```
bhgomes, line 385: Maybe leave a comment on why you have to collect before looping.

bhgomes (changes requested):

manta-pay/src/simulation/mod.rs
SupremoUGH: `load_accounts` now calls `update_authorization_context` internally.

bhgomes (approved):

zqhxuyuan (approved):

PR #347, opened 2023-05-18
Signer bug fix

SupremoUGH:
Fixes the infamous Wallet `AssetSpent` bug. Goes together with https://github.com/Manta-Network/sdk/pull/129

---


manta-accounting/src/wallet/signer/functions.rs
```diff
@@ -150,7 +149,7 @@ fn insert_next_item<C>(
     assets: &mut C::AssetMap,
     parameters: &Parameters<C>,
     identified_asset: IdentifiedAsset<C>,
-    nullifiers: &mut Vec<Nullifier<C>>,
+    nullifiers: &mut C::NullifierMap,
```
ghzlatarev, line 152: one thing im noticing is that the map doesn't get cleared when doign a resync.
not sure if that's a bad thing though

manta-accounting/src/wallet/signer/nullifier_map.rs
```diff
@@ -0,0 +1,125 @@
...
+#[cfg(feature = "std")]
+use {core::hash::Hash, std::collections::HashSet};
+
+/// Nullifier Map
+pub trait NullifierMap<T>: Default {
+    /// Creates a new [`NullifierMap`].
+    fn new() -> Self;
+
+    /// Inserts `item` in `self`.
+    fn insert(&mut self, item: T) -> bool;
```
ghzlatarev, line 30: i'd also add len() function in case we need to debug in the future, like in `wasm_debugging_utilities`

manta-accounting/src/wallet/signer/functions.rs
ghzlatarev: nvm they do get reset

manta-accounting/src/wallet/signer/functions.rs
```diff
@@ -198,11 +192,7 @@ where
 {
     let (_, utxo, nullifier) =
         parameters.derive_spend(authorization_context, identifier, asset.clone(), rng);
-    if let Some(index) = nullifiers
-        .iter()
-        .position(move |n| n.is_related(&nullifier))
-    {
-        nullifiers.remove(index);
+    if nullifiers.contains_item(&nullifier) {
```
ghzlatarev, line 195: do you mind writing out the logic of why we mutate the assets/nullifiers here once and then at the end as well, to make it easier for new people to get into this part of the code.

manta-accounting/src/wallet/signer/functions.rs
ghzlatarev: actually maybe it's a good idea to release all the nullifiers that are not my own
it might get laggy to carry around 100s of thousand of nullifiers in memory

manta-accounting/src/wallet/signer/functions.rs
ghzlatarev, line 1: this is not enough yet, even though the nullifiers are persistent, they are still being pulled in parallel with the utxos.
We need to first pull all the nullifiers and decouple this function, in order to not miss any utxo checks.

manta-accounting/src/wallet/signer/functions.rs
SupremoUGH: so we insert the assets as non-provable straight away in the merkle tree, instead of inserting them as provable and then removing the proof.

manta-accounting/src/wallet/signer/functions.rs
SupremoUGH: I think it should be enough. Note that the `assets.retain` part of the code is always being called, so in the final batch of the `pull` function no asset will be left unchecked against the nullifiers.

manta-accounting/src/wallet/signer/nullifier_map.rs
SupremoUGH: good idea

manta-accounting/src/wallet/signer/functions.rs
SupremoUGH: It is time vs space, as usual. If we release them from memory, then next time we sync we have to: pull them all again and insert them all (which is O(n) where n is the total number of nullifiers vs O(m) where m is the number of new nullifiers).

A more interesting question is: can we just keep the Nullifiers that belong to us? I think we can by using the OutgoingNotes but I didn't want to overcomplicate this PR. That's a further optimization for when we are sure the bug is fixed.

manta-accounting/src/wallet/signer/functions.rs
ghzlatarev: Hmm yeah you're right.
Weirdly though Daniel is still getting the error somehow on production.
On the local testnet i can reliably not reproduce the error, but there the nullifiers are not that many and are pulled in the first go.

manta-accounting/src/wallet/signer/functions.rs
SupremoUGH: https://github.com/Manta-Network/manta-rs/issues/348

ghzlatarev (approved):

Apokalip (approved):

PR #353, opened 2023-06-05
Restore Merkle tree pruning for the wallet

SupremoUGH:
Reenables the Merkle tree pruning functionality for the wallet, this time as a BTreeMap instead of vec, which allows for almost 100% saved space (as opposed to 50%). Fixes some things in `fork` which were incompatible with pruning, as well as some indexing issues which were incompatible with the branches. 

Should also speed up sync time because of the batch optimization when merging a branch into its trunk. Not now, see the issue: https://github.com/Manta-Network/manta-rs/issues/354

Goes together with: https://github.com/Manta-Network/sdk/pull/133

# Note for reviewers

When reviewing, please make sure that the indexing is right, meaning: indexes in the Partial Merkle tree and indexes in its leaf map are off by `self.starting_leaf_index()`. Make sure each function is using the right index, depending on whether they are interacting with the tree or just with its `LeafMap`.

---

### DIFF
--- a/manta-crypto/src/merkle_tree/leaf_map.rs
+++ b/manta-crypto/src/merkle_tree/leaf_map.rs
@@ -138,7 +138,7 @@ where
 #[derivative(
     Clone(bound = "LeafDigest<C>: Clone"),
     Debug(bound = "LeafDigest<C>: Debug"),
-    Default(bound = "LeafDigest<C>: Default"),
+    Default(bound = ""),
     Eq(bound = "LeafDigest<C>: Eq"),
     Hash(bound = "LeafDigest<C>: Hash"),
     PartialEq(bound = "LeafDigest<C>: PartialEq")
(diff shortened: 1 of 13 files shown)


Apokalip (dismissed):
lgmt, I dont know if its not a bad idea to have tests for the leaf indexes(reviewer note)

manta-crypto/src/merkle_tree/partial.rs
```diff
@@ -344,6 +361,44 @@ where
...
+        M: Default,
+        InnerDigest<C>: Clone + Default + PartialEq,
+        LeafDigest<C>: Clone + Default,
+    {
+        let marked_leaf_digests = marked_leaf_digests.into_iter();
+        if matches!(marked_leaf_digests.size_hint().1, Some(max) if max <= capacity::<C, _>() - self.len())
+        {
+            for (marking, leaf_digest) in marked_leaf_digests {
+                if marking {
+                    assert!(self.push_digest(parameters, || leaf_digest),
```
zqhxuyuan, line 390: why not error instead of assert?

manta-crypto/src/merkle_tree/leaf_map.rs
```diff
@@ -545,7 +581,7 @@ where
     fn remove(&mut self, index: usize) -> bool {
         match self.last_index {
             Some(current_index) if index == current_index => false,
-            _ => !matches!(self.map.remove(&index), None),
+            _ => self.map.remove(&index).is_none(),
```
zqhxuyuan, line 584: The old one using `!matches!`, I think this need to be `!self.map.remove(&index).is_none()`

SupremoUGH: Yes, you're right. Good catch!

Apokalip (approved):

zqhxuyuan (approved):

PR #355, opened 2023-06-06
Consolidate utxos

SupremoUGH:
Adds the `consolidate` method to the signer, which merges several UTXOs into one with the total value. 
To be merged after: https://github.com/Manta-Network/manta-rs/pull/353
Goes together with: https://github.com/Manta-Network/sdk/pull/133

---

### DIFF
--- a/manta-accounting/src/wallet/signer/functions.rs
+++ b/manta-accounting/src/wallet/signer/functions.rs
@@ -59,6 +59,8 @@ use manta_util::{
     vec::VecExt,
 };
 
+use super::ConsolidationRequest;
+
 /// Returns the default account for `accounts`.
 #[inline]
 pub fn default_account<C>(accounts: &AccountTable<C>) -> Account<C::Account>
@@ -511,6 +513,34 @@ where
     })
 }
 
+///
+#[inline]
+fn custom_select<C>(
+    accounts: &AccountTable<C>,
+    assets: &C::AssetMap,
+    parameters: &Parameters<C>,
+    request: &ConsolidationRequest<C>,
+    rng: &mut C::Rng,
+) -> Result<Selection<C>, SignError<C>>
+where
+    C: Configuration,
+    IdentifiedAsset<C>: PartialEq,
+{
+    if !request.check_consolidation_request(assets) {
+        return Err(SignError::InvalidConsolidationRequest);
+    }
+    let selection = request.select::<C::AssetMap>();
+    Selection::new(selection, move |k, v| {
+        Ok(build_pre_sender::<C>(
+            accounts,
+            parameters,
+            k,
+            Asset::<C>::new(request.id().clone(), v),
+            rng,
+        ))
+    })
+}
+
 /// Builds a [`TransferPost`] for the given `transfer`.
 #[inline]
 fn build_post_inner<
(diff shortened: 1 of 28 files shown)


manta-accounting/src/wallet/signer/mod.rs
```diff
@@ -835,6 +846,217 @@ where
...
+        let asset_id = &assets
+            .first()
+            .expect("This cannot fail because of the check above.")
+            .asset
+            .id;
+        (!(1..number_of_assets).any(|i| {
+            assets[i..].contains(&assets[i - 1])
+                && assets[i].asset.id.ne(asset_id)
+                && assets[i].asset.value.eq(&Default::default())
+        }))
```
ghzlatarev, line 986: This won't return None when only one of these conditions is met, which seems sus.

manta-accounting/src/wallet/signer/functions.rs
```diff
@@ -1039,6 +1132,33 @@ where
...
+{
+    let result = consolidate_internal(
+        parameters,
+        accounts.ok_or(SignError::MissingSpendingKey)?,
+        assets,
+        utxo_accumulator,
+        request.try_into()?,
+        rng,
+    )?;
+    utxo_accumulator.rollback();
```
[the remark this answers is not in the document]
SupremoUGH: To generate the transaction, you need to create receivers that you may need to spend again, so you have to update the internal UTXO accumulator. But that doesn't correspond to the actual ledger state (e.g. someone else may have minted some UTXOs while you were generating the proofs or your transaction may be rejected) so you rollback to the original state after signing. Only after syncing you want to update the accumulator in a definite manner (calling `commit` instead) because that's actual ledger data that you're importing.

manta-accounting/src/wallet/signer/mod.rs
SupremoUGH: Good catch

manta-accounting/src/wallet/signer/mod.rs
```diff
@@ -835,6 +846,217 @@ where
...
+            .0
+            .first()
+            .expect("Consolidation requests must have at least 2 elements")
+            .asset
+            .id;
+        let asset_map_assets = asset_map
+            .asset_vector_with_id(id)
+            .into_iter()
+            .map(|(identifier, asset)| IdentifiedAsset::<C>::new(identifier, asset))
+            .collect::<Vec<_>>();
```
ghzlatarev, line 1029: How big is this Vector ? Maybe we should collect it into a hash set because of the following lookup

SupremoUGH: The vector shouldn't be too big. Also, `HashSet` is "forbidden" for wallet implementations (I got some trouble with the WASM part of the sdk when there were `std` calls).

ghzlatarev: could write our own i guess lol (in my free time maybe)

manta-pay/src/test/signer.rs
```diff
@@ -281,3 +281,63 @@ async fn find_the_bug() {
         wallet.signer_mut().prune();
     }
 }
+
+/// Tests [`consolidate`](manta_accounting::wallet::Wallet::consolidate) works as expected.
+#[ignore] // We don't run this test on the CI because it takes a long time to run.
+#[tokio::test]
+async fn consolidation_test() {
```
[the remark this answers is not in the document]
SupremoUGH: I didn't add them because this is super-safe already. If you somehow managed to create an invalid consolidation request the resulting transfer would be invalid, e.g., you try to spend the same asset twice (duplicates in the request), you try to spend assets with different IDs (transfer protocol always rejects that), you try to spend assets you don't own (the `check` function returns `false`).
The only ones that would pass are those which involve zero assets (which besides being stupid and costing gas won't do any harm, and why would you select zero assets in the first place?), including the consolidation request with a single element.

Apokalip (approved):

